PR #4, opened 2016-02-16
Someone on the internet is using XCB poorly

stricherkind:
Avoid unnecessary round-trip. See https://xcb.freedesktop.org/tutorial/.


SaschaWillems:
With glfw now having Vulkan support I plan on dropping the OS specific stuff anyway. It's the first time I used XCB, so please bear with me.

PR #7, opened 2016-02-16
Changed to & operator

kretash:


vtlmks:
Why? && is fine there I think; if formatProps.optimalTilingFeatures is not true, then && makes an early exit.

shurcooL:
Why did you close this PR? Is it invalid?

kretash:
I don't really understand what this part is trying to do. At first its seamed like a typo, but I'm not sure anymore.

PR #8, opened 2016-02-16
Fix needed for outsource build

akazantsev:
Couldn't build samples in another dir without this fix


kainino0x:
Confirming that this fix works for me.

PR #10, opened 2016-02-17
Fix crash by setting valid physical device enumeration parameter

zhenyw:
If pPhysicalDevices is not NULL for vkEnumeratePhysicalDevices(),
have to properly set pPhysicalDeviceCount.

This fixed crash on intel vulkan driver.


SaschaWillems:
I have pushed a more elegant fix with 40f13fc770bc714235c0545248d72aab967ba314

The example base will now enumerate the number of available devices, put them in a vector and select the first device supporting Vulkan.

PR #22, opened 2016-02-18
Fix CMake project name

encelo:
The CMake project name is set to be the same as the name of the first example.
I suggest it gets changed to the name of the Visual Studio solution included in the repository.


SaschaWillems:
Thx!

PR #26, opened 2016-02-19
Fix building on Linux

janisozaur:
This finds and links assimp properly now

The `cmake/assimp-config.cmake` file is taken from Arch's [`assimp` package](https://www.archlinux.org/packages/community/x86_64/assimp/)


cmake/assimp-config.cmake
```diff
@@ -0,0 +1,75 @@
...
+#set(Boost_USE_STATIC_RUNTIME OFF)
+#find_package(Boost ${ASSIMP_Boost_VERSION} EXACT COMPONENTS thread date_time)
+#if(Boost_VERSION AND NOT "${Boost_VERSION}" STREQUAL "0")
+#	set( ASSIMP_INCLUDE_DIRS "${ASSIMP_INCLUDE_DIRS}" ${Boost_INCLUDE_DIRS})
+#else(Boost_VERSION AND NOT "${Boost_VERSION}" STREQUAL "0")
+#	message(WARNING "Failed to find Boost ${ASSIMP_Boost_VERSION} necessary for assimp")
+#endif(Boost_VERSION AND NOT "${Boost_VERSION}" STREQUAL "0")
+
+# the boost version assimp was compiled with
+set( ASSIMP_Boost_VERSION "1.59")
```
1ace: I don't know if this line is required (don't know enough about either cmake or assimp), but it looks weird to have this static value there

janisozaur: Yes, I agree, it all looks slightly suspicious. I'll try replacing this file with one from upstream: https://github.com/assimp/assimp/blob/master/cmake-modules/Findassimp.cmake

1ace:
I think this can be closed?

SaschaWillems:
Due to the recent changes in the CMakeLists.txt this should work fine on Linux now.

PR #30, opened 2016-02-20
Cygwin """Support"""

Mr-Quiggles:
For whatever reason `CMakeLists.txt` didn't properly diff; all that's added were:

```
# Treat Cygwin as Windows, not Linux (it's easier this way)
IF(CYGWIN)
    set(WIN32 "1")
ENDIF(CYGWIN)
```

and

```
# CMake fails to find the vulkan library; hard coded
IF(CYGWIN)
    set(VULKAN_LIB ${CMAKE_SOURCE_DIR}/libs/vulkan/vulkan-1.lib)
ENDIF(CYGWIN)
```

before and after the first `IF(WIN32)` block.

Also supplied is a copy of `assimp` compiled with `x86_64-w64-mingw32-gcc` (`MinGW64`). There was nagging about missing `libassimp.dll` (I'll assume since there's already a supplied binary compiled with Visual Studios, a copy compiled with MinGW **should** be supplied as well).
Contrarily, I didn't supply `libgcc_s_seh-1.dll` and `libstdc++6.dll` for a number of reasons (they're easily attainable from your development environment, they're only for MinGW and not Cygwin-based gcc compilers, within MinGW itself, there's several exception handling types (`seh`, `sjlj`, `dw2` handling all use different dlls).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,6 +12,10 @@ include_directories(external/gli)
 include_directories(external/assimp)
 include_directories(base)
 
+IF(CYGWIN)
+	set(WIN32 "1")
+ENDIF(CYGWIN)
+
 IF(WIN32)
 	find_library(VULKAN_LIB NAMES vulkan-1 vulkan PATHS ${CMAKE_SOURCE_DIR}/libs/vulkan)
 	find_library(ASSIMP_LIB NAMES assimp PATHS ${CMAKE_SOURCE_DIR}/libs/assimp)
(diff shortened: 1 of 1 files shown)


pushrax:
You converted all the line endings, perhaps submit without that.

SaschaWillems:
I just wanted to test this on cygwin and added the changes from this PR to get cmake to work, but I don't think that's enough as it complains about XCB functions so I guess there is a missing define. 

Could you check again with a current version of cmake and cygwin and do a proper PR that only contains the changed lines?

SaschaWillems:
Closing due to inactivity.

PR #38, opened 2016-02-22
Fixed a mistake in sem.frag: pow(x,y) is undefined when x < 0

jaakkowork:
See https://www.khronos.org/registry/spir-v/specs/1.0/GLSL.std.450.html

"Pow

Result is x raised to the y power; x^y. Result is undefined if x < 0. Result is undefined if x = 0 and y ≤ 0."


SaschaWillems:
Thanks! The example now looks like it's supposed to look an AMD too :+1:

PR #49, opened 2016-02-27
fix descriptor pool creation in computeshader demo

cwabbott0:
Set the right number of descriptors to fixes an out-of-memory error on the Intel driver on the latest mesa.


SaschaWillems:
Thx, I fixed this with commit 3b27fe6fc2af95a4c8867267b7c988552f7a215c that also fixes a crash when trying to switch between the compute pipleines.

PR #55, opened 2016-03-02
fix handle_event -> handleEvent in shadowmapping

ChristophHaag:
The shadow mapping example didn't build for me:
```
shadowmapping/shadowmapping.cpp:1177:18: Fehler: »class VulkanExample« has no member named »handle_event«
```
In all the other examples it's handleEvent instead of handle_event and that works.

For some reason the github interface seems to display the commit wrong, maybe, because of the line endings?


SaschaWillems:
Thx!

PR #59, opened 2016-03-05
Update README.md

dustin-biser:
Fixed typo.  "scree space" -> "screen space".


jeffro256:
Hey @SaschaWillems, not be to rude, but I think that this should be merged. It's a very small change, but correct!

PR #63, opened 2016-03-06
Remove redundant assignment to vertex attribute binding

pushrax:
The binding is set to `VERTEX_BUFFER_BIND_ID` a few lines up, so this assignment is redundant (and potentially wrong if the definition changes).


pushrax:
Found an unused variable across a few files; figured I may as well remove them.

SaschaWillems:
Thx! I have made some changes recently, mostly to the barriers used in the examples, so I won't be able to merge this. I'll also take a look a the unused variables.

Issue #42, opened 2016-02-23
Semaphore must be created with 0 flags


httpdigest:
Sorry for posting so many things here all the sudden, but I am porting your very great examples to Java/LWJGL and everything I encounter I would like to have fixed in your examples, too. :)

[This line](https://github.com/SaschaWillems/Vulkan/blob/master/triangle/triangle.cpp#L209) should use 0 as flags. [The spec says](https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/xhtml/vkspec.html#VkSemaphoreCreateInfo) that this must be zero and is reserved for future use.

I think the fence flags only apply to fences and not to semaphores.

SaschaWillems:
> Sorry for posting so many things here all the sudden, but I am porting your very great examples to Java/LWJGL and everything I encounter I would like to have fixed in your examples, too. :)

No excuses necessary ;) I'm pretty sure that there are lots of small bugs left from using earlier (unfinished) Vulkan SDK versions and specs.

httpdigest:
Aright, cool! :) I want to go over all of your demos in the next days, and see if there might be some more minor things. Currently, your demos are the de-facto goto place for learning Vulkan by reading source code. It should be absolutely perfect. :)

Maybe one question: You are some "drain queue" waits with vkQueueWaitIdle, [for example here](https://github.com/SaschaWillems/Vulkan/blob/master/triangle/triangle.cpp#L286).

Do you know whether this is actually necessary, when also waiting on the vkAcquireNextImageKHR semaphore when beginning the next render pass?

I don't want to bother you with specific questions about Vulkan, which cannot also be read on in the spec, but truth is, the spec is a very bad place to go to when trying to learn about synchronization/ordering/fences/semaphores stuff, and where this is necessary and when exactly, especially with the WSI extensions.
Do you know whether there is already a good programming guide? Most of the stuff I know already about Vulkan I actually learnt from the [Mantle programming guide](https://www.amd.com/Documents/Mantle-Programming-Guide-and-API-Reference.pdf).

Thanks!

SaschaWillems:
> Maybe one question: You are using some "drain queue" waits with vkQueueWaitIdle, for example here.

> Do you know whether this is actually necessary, when also waiting on the vkAcquireNextImageKHR semaphore before beginning the next render pass?

No, I think they're no longer necessary with the semaphore in place. They're probably left overs from the early development phase.

[...]

PR #64, opened 2016-03-06
Don't set flags on VkSemaphoreCreateInfo

pushrax:
`VkSemaphoreCreateInfo`: [`flags` is reserved for future use](https://vulkan.lunarg.com/v1/docs/v1.0.3.1/vulkandocs/vkspec.html#VkSemaphoreCreateInfo)

### DIFF
--- a/shadowmapping/shadowmapping.cpp
+++ b/shadowmapping/shadowmapping.cpp
@@ -526,8 +526,7 @@ public:
 	{
 		VkResult err;
 		VkSemaphore presentCompleteSemaphore;
-		VkSemaphoreCreateInfo presentCompleteSemaphoreCreateInfo =
-			vkTools::initializers::semaphoreCreateInfo(VK_FLAGS_NONE);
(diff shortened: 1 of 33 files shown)


shadowmapping/shadowmapping.cpp
```diff
@@ -526,8 +526,7 @@ class VulkanExample : public VulkanExampleBase
 	{
 		VkResult err;
 		VkSemaphore presentCompleteSemaphore;
-		VkSemaphoreCreateInfo presentCompleteSemaphoreCreateInfo =
-			vkTools::initializers::semaphoreCreateInfo(VK_FLAGS_NONE);
+		VkSemaphoreCreateInfo presentCompleteSemaphoreCreateInfo = vkTools::initializers::semaphoreCreateInfo();
```
pushrax: GH is showing this diff incorrectly, for the actual change see https://github.com/SaschaWillems/Vulkan/pull/64.patch

SaschaWillems:
Thx, this was already noted in #42 and I already started changing this in a few examples locally, using VK_FLAGS_NONE instead of VK_FENCE_CREATE_SIGNALED_BIT.

PR #67, opened 2016-03-07
Fix some error checks in triangle.cpp

Semmy13:
In VulkanExample::prepareVertices:
 - the return code for vkAllocateMemory (for the vertex buffer) wasn't cheked;
 - the assert after vkUnmapMemory is not required (vkUnmapMemory doesn't return a VkError)


SaschaWillems:
Thx! The assert for vkUnmapMemory was a left over from an older SDK version.

PR #76, opened 2016-03-10
Computeshader fixes

jekstrand:
This gets the computeshader demo working on the Intel Linux driver

### DIFF
--- a/base/vulkanTextureLoader.hpp
+++ b/base/vulkanTextureLoader.hpp
@@ -154,7 +154,9 @@ namespace vkTools
 
 				// Setup texture as blit target with optimal tiling
 				imageCreateInfo.tiling = VK_IMAGE_TILING_OPTIMAL;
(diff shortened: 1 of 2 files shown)


SaschaWillems:
Thx for this :+1:  
I'll test this on my Intel machine later this week. One thing I don't like to do though is adding the VK_IMAGE_USAGE_STORAGE_BIT for every texture, I think I'll go with an overload there where you can pass additional parameters to the loadTexture function

jekstrand:
That seems like a reasonable thing to do.  Thanks for taking a look at this.  The Intel Linux driver is very sensitive to having the right usage bits set on buffers and images and using the correct descriptor types.

SaschaWillems:
Commited your changes with a few additions (like passing the usage flag bits when loading the texture) and this example now finally works on my Ultrabooks integrated skylake GPU.

Thanks again for this :+1:

Issue #33, opened 2016-02-21
Half of the samples crash


cheery:
* bloom
 * deferred
 * displacement
 * geometryshader
 * instancing
 * mesh
 * occlusionquery
 * offscreen
 * parallaxmapping
 * pushconstants
 * radialblur
 * shadowmap
 * skeletalanimation
 * sphericalenvmapping
 * tesselation
 * texturecubemap
 * vulkanscene

GTX 960, Ubuntu 14.10

cheery:
[vulkanscene.dump.txt](https://github.com/SaschaWillems/Vulkan/files/139402/vulkanscene.dump.txt)

Here's a vulkanscene dump. Maybe it tells you why it crashes and where?

CapsAdmin:
Same thing happens to me using a gtx 970 on kubuntu 15.10 with neon packages (so newer Qt if that matters). I also had to install kernel 4.2.0-29-generic in order to use the driver.

If it helps here's deferred's vktrace [deferred.vktrace.zip](https://github.com/SaschaWillems/Vulkan/files/139517/deferred.vktrace.zip)
It doesn't crash when replaying it though.

SaschaWillems:
Did you compile from source or are you using the prebuild binaries? Can you give me a call stack? From the list it seems that only samples using models (most of them) crash, so they may either not find the data folder or something is wrong with ASSIMP. If you compliled yourself, what ASSIMP version did you use?

CapsAdmin:
I compiled everything from source with `cmake -DCMAKE_BUILD_TYPE=Debug ..`. 
This is an out of source build but I'm aware that the working directory needs to be Vulkan/bin/

If I attempt to run bloom outside of this working directory it will error like this
`bloom: /media/caps/ssd_840_120gb/Vulkan/base/vulkanTextureLoader.hpp:360: void vkTools::VulkanTextureLoader::loadCubemap(const char*, VkFormat, vkTools::VulkanTexture*): Assertion `!texCube.empty()' failed.`

Which is expected.

The "texture" sample manages to load the texture fine though. (Unrelated but the plane looks like it has a lod level is missing (guessing) cause it has a green shade in some angles)

SaschaWillems:
Can you please run the demos with validation layers enabled and check for errors?

CapsAdmin:
I set ENABLE_VALIDATION to true in bloom.cpp

```
[...]

PR #81, opened 2016-03-14
Fix building on linux (and possibly on windows)

dcbaker:
This pull corrects the build on linux in a couple of ways:

First it stops falling back to a dll.a file, which I assume was actually meant as a fallback for windows

Second, It uses find_package with the config option for assimp. This config file should not be imported (unlike what is suggested by #26), it is provided by the upstream and if assimp is installed the cmake-config should be properly installed and available on the system. This should be faster and more robust than the current find_libraries approach.


SaschaWillems:
Thx!

CapsAdmin:
This works for me on kubuntu 15 as well. I had to do "export ASSIMP_DIR=*path to assimp git repo*" though.

dcbaker:
Now that I look back at it, I think that the first patch might be wrong. ld gave me an error about libass.dll.a that made me think it wasn't a linux object, but it is a .a file?

@CapsAdmin, did you have to do that before or after my patches?

CapsAdmin:
Compiling never worked for me before. See #33

SaschaWillems:
Yes, libassimp.dll.a is not a linux file and was a mistake. I used this to check if all projects compile fine on Winodws with MinGW.

SaschaWillems:
Sadly this won't work for me on linux anymore. It can't locate assimpd, even though I installed libssimp-dev.

I'm on Ubuntu. Any ideas? I'd prefer having a CMakeLists.txt that works out of the box :/

dcbaker:
hmmmm, it appears that for some reason ubuntu, (or maybe debian?) doesn't install the cmake-config files. Let me do some more digging, at the worst I'll write a FindLibassimp.cmake that does the proper falling back.

PR #85, opened 2016-03-17
More assimp build fixes on linux

cwabbott0:
This makes things better for distros like Fedora that package assimp 3.1, by making it more obvious that assimp 3.2 is required and making overriding the system assimp easier.


SaschaWillems:
This should be obsolete due to recent changes in CMakeLists.txt, except for the version required (not sure if that's a good idea).

PR #87, opened 2016-03-18
Cleaned macros in VulkanBaseExample::VulkanBaseExample(bool)

jeffro256:
I made the macro in VulkanBaseExample::VulkanBaseExample(bool) which "checks for validation command line flag" cleaner. It had 2 ifs that could be merged into one, and this new one is easier on my eyes. :)

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -475,9 +475,7 @@ VulkanExampleBase::VulkanExampleBase(bool enableValidation)
 			enableValidation = true;
 		}
 	}
-#endif
-
-#ifndef _WIN32
+#else
 	initxcbConnection();
 #endif
 	initVulkan(enableValidation);


1ace:
Those two blocks do entirely different things: one enables the validation layer, while the other connects to a window system. IMHO those should not be merged into a single block.

What could be useful though is to add a comment before `initxcbConnection();` to explain what it's doing (connect to a WS).

That second block might also benefit from a better test than "not windows" tough, such as `#ifdef __unix__` (or `__linux__` if `__APPLE__` is to be excluded (does XCB support XQuartz? I don't know how conformant it is)).

SaschaWillems:
@1ace : That's on my todo list. Apple is not an option yet, though MetalVK may be a target at some point in the future. But I'll at least want to add Android support for the examples (removing the need for separate android samples), so I'll have to address the ifdef blocks at some point in the future anyway.

jeffro256:
@1ace Yeah, after looking back over that code, I can see why that decision was made. I agree that there should be comments on `initxcbConnection()`, I had to comb through Google to understand how this worked, as I had no idea what XCB was beforehand. For my projects, I had used GLFW.

PR #88, opened 2016-03-19
Replace 'not Windows' with 'Linux' test

1ace:
A first step :]

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -477,9 +477,11 @@ VulkanExampleBase::VulkanExampleBase(bool enableValidation)
 	}
 #endif
 
-#ifndef _WIN32
+	// Connect to the window system
+#ifdef __linux__
 	initxcbConnection();
 #endif
+
 	initVulkan(enableValidation);
 	// Enable console if validation is active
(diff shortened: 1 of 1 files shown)


1ace:
Actually, diving a bit into the code, I see it's full of `#ifdef _WIN32 /*…*/ #else /* assume linux */ #endif`. Those should all be also rewritten into `#elif defined(__linux__)` to be consistent.
By the way, would you be open to a rewrite of all the `#ifdef FOO` into `#if defined(FOO)`? This latter form is much more flexible, and would be more consistent with those `#elif defined(BAR)` cases.

This PR is therefore a bit pointless, but I'll leave it open if you still want it.
Feel free to just close it and do things yourself though.

SaschaWillems:
> By the way, would you be open to a rewrite of all the #ifdef FOO into #if defined(FOO)? This latter form is much more flexible, and would be more consistent with those #elif defined(BAR) cases.

Yes. Rewriting the defines is on my todo list anyway as I also want to add Android support to all the demos, that requires are few more (and cleaner) defines. I'll hope to get started with the Android stuff next weekend and I'll loo into the other defines then as well.

1ace:
See #91

PR #91, opened 2016-03-20
Fix ifdef

1ace:
- Rewrite every `#ifdef FOO` as `#if defined(FOO)`  
  Leave the `#ifndef` alone though.
- Fix OS detection  
  Everything that is not Windows isn't necessarily Linux :]

When adding Android support, be careful to add `#elif defined(__ANDROID__)` before `__linux__`, as Android is also a Linux :+1:

### DIFF
--- a/android/computeparticles/computeparticles.NativeActivity/android_native_app_glue.h
+++ b/android/computeparticles/computeparticles.NativeActivity/android_native_app_glue.h
@@ -26,7 +26,7 @@
 #include <android/looper.h>
 #include <android/native_activity.h>
 
-#ifdef __cplusplus
+#if defined(__cplusplus)
 extern "C" {
 #endif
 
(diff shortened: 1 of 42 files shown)


SaschaWillems:
Thanks for your effort :+1:  

I didn't think about using #elif in the first place, but this is def the better way.

Though I already started adding android support for the examples, which means I already changed some of the defines, so I don't think I'll be able to merge this pull request, but rather pick the changes that apply.

1ace:
Sure, whatever works for you :+1:

You can also [pull it on your computer](https://help.github.com/articles/checking-out-pull-requests-locally/) and merge it there, where it's easy to handle any conflict.

SaschaWillems:
Closing this due to the conflicts. I did replace lots of #ifdefs with #if defines and I'll replace existing ones as I see fit.

PR #93, opened 2016-03-20
Fix circular/modulo arithmetic

1ace:
The input param was also unsigned, leading to an overflow when passing `-1`.


SaschaWillems:
Fixed manually with https://github.com/SaschaWillems/Vulkan/commit/8fbfb7c22bd6f9fa1532e45db8e6d0f5434f6e60

PR #94, opened 2016-03-20
Use `uint32_t` to store size()

1ace:
I didn't touch the `size_t`, as it's even better IMHO.
`uint32_t` was the most used for `size()` in your code, so I went with that for everything else.

### DIFF
--- a/android/computeparticles/computeparticles.NativeActivity/computeparticles.cpp
+++ b/android/computeparticles/computeparticles.NativeActivity/computeparticles.cpp
@@ -599,7 +599,7 @@ public:
 
 		VkResult err;
 
-		for (int32_t i = 0; i < drawCmdBuffers.size(); ++i)
+		for (uint32_t i = 0; i < drawCmdBuffers.size(); ++i)
 		{
 			// Set target frame buffer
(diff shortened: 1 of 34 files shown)


Voultapher:
When changing it anyways, what is motivation of not using size_t?

1ace:
@Voultapher: that's literally the only other message in this thread; did you read it? :wink:

Voultapher:
>uint32_t was the most used for size() in your code, so I went with that for everything else.

sure, but afaik using uint32_t would be technically wrong and by that a bug that should be addressed. As this pull request is about changing this behavior, why not switch every size() storing variable to size_t, the way it is meant to be used.

GoldSciences:
Actually using size_t is a source of bugs as it builds as its meaning changes from signed to unsigned and its size in bytes also changes, which is very risky for things like boundary checks. Using such an unreliable type as parameter for interfaces doesn't provide any benefit and the problems it causes are well known. There are other reasons that comes to mind such as -1 not having the same bit pattern for uint32_t as for int64_t, which is another source of bugs that may go unnoticed.

PR #95, opened 2016-03-20
Remove dead code

1ace:
DO NOT MERGE!

OK, now that I got your attention… :grin:
I expect a few of those might be sign that you meant to do something else, and might have used the wrong variable at some place instead of the one you just created, or things like that.
Most of them are probably just copy/paste leftovers.

This is to be considered as a list of things to double check. Once you are sure it's all good, go ahead and merge.
And if some of those were indeed meant to  be used, I'll gladly fix the pull request :wink:


SaschaWillems:
Many of these have been fixed with recent commits, the rest has been fixed with https://github.com/SaschaWillems/Vulkan/commit/0c1c89c3cb3a0ebfcdc4c43ebf6b6c2efda9874b.

Issue #96, opened 2016-03-20
Missing virtual destructor in VulkanExample


1ace:
In every example, you create a global `vulkanExample` object to handle events, and then delete it as soon as the loop closes. When calling `delete(vulkanExample);`, gcc complains:

> warning: deleting object of polymorphic class type ‘VulkanExample’ which has non-virtual destructor might cause undefined behaviour [-Wdelete-non-virtual-dtor]

My polymorph-fu is weak (I've been mostly coding drivers in C lately), but I think you're missing a virtual destructor to be overridden when subclassing, to handle whatever the subclass adds.

I'm working on a fix that makes `vulkanExample` local to `main()` (it works, but needs to be applied to every example).
I'm not going to push that tonight: It's 5am, I should really to go sleep :P

Edit: Still not asleep, but awake enough to realize I've been saying something dumb: my "fix" doesn't fix that issue, it simply avoids allocating stuff on the stack. Going to sleep now.

PR #99, opened 2016-03-20
Move VulkanExample allocation to the stack

1ace:
As I mentioned in #96, this is a rewrite of all the `VulkanExample` allocations onto the stack.

The pattern applied to all those changes is:
- Move allocation to the stack (`vulkanExample = new VulkanExample();` → `VulkanExample vulkanExample;`)
- Rename global object (`VulkanExample *vulkanExample;` → `VulkanExample *vulkanExampleGlobal;`)
- Give global pointer to local object (`vulkanExampleGlobal = &vulkanExample;`)

It does silence the compiler warning mentioned in #96, but I'm not sure this actually fixes the issue, so it should still be addressed.

Also, [don't hesitate to just close any of my PR if they don't align with what you want to do](https://github.com/blog/2124-kindly-closing-pull-requests) :wink:

### DIFF
--- a/bloom/bloom.cpp
+++ b/bloom/bloom.cpp
@@ -1303,27 +1303,27 @@ public:
 	}
 };
 
-VulkanExample *vulkanExample;
+VulkanExample *vulkanExampleGlobal;
 
 #ifdef _WIN32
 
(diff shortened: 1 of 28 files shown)


SaschaWillems:
Closing this due to the (not so) recent changes of changing the main entry points into centralized macros.

PR #100, opened 2016-03-22
Replace external libraries with proper git submodules

1ace:
See https://git-scm.com/book/en/v2/Git-Tools-Submodules for a primer on the use of `git submodule`.

Tl;dr:
- When cloning (first time), add `--recursive` to automatically handle submodules
- If you have already cloned the repo and are updating past this commit, run:
  
  ```
  $ git submodule init
  $ git submodule update
  ```

For `gli`, I used the `0.7.0.0` tag as this is the version you seemed to be using.
Note that you'll need to rewrite a lot of `gli::*` stuff if you ever want to upgrade to `0.8.x.y`.


SaschaWillems:
Thx :+1: But let me check this before merging. I initially wanted to use submodules, but didn't do so because (at least during early development) I've had troubles with the most recent versions of gli (which should now be fine) and assimp. 

But yes, submodules are the way to go. I'll double check if everything builds and runs fine with your commit and merge if it's okay.

1ace:
As I mentioned, I didn't use the latest version of `gli` but the tag `0.7.0.0` instead. I expect this is the "troubles" you are talking about.

SaschaWillems:
gli and glm are now proper submodules. ASSIMP is a tad to big for a submodule for my taste and esp. on Android had some troubles it'll stay as a copy.

PR #101, opened 2016-03-22
Remove VK_PROTOTYPES define

1ace:
This guard macro was inverted shortly before the release; use VK_NO_PROTOTYPES when you do *not* want the prototypes.
Here, we clearly *do* want the prototypes, so we don't need anything to be defined.

I also just noticed that the documentation is wrong: KhronosGroup/Vulkan-Docs#150


1ace:
I haven't tested the Windows part of this patch as I don't have Vulkan-capable hardware at home, but it looked like this was the right format, so I guess it should work :stuck_out_tongue_closed_eyes:

SaschaWillems:
Thx, another left over from early development.

Issue #102, opened 2016-03-22
handleEvent() is defined but not used


1ace:
> warning: ‘void handleEvent(const xcb_generic_event_t*)’ defined but not used [-Wunused-function]

Is your event handling working correctly? From what the compiler says and what I've seen by looking at the code, it seems like it would always be ignored.
I don't have Vulkan-capable hardware at home, so I can't test it myself.

SaschaWillems:
LInux event handling is not yet finished, see #56. Common event handling (zoom, rotate, etc.) is working, but specific example keys not.

1ace:
Right, sorry. I should've checked the existing issues :S

PR #103, opened 2016-03-22
Add some debug warnings

### LINKED ISSUE #90: Bad `width` test
In base/vulkanswapchain.hpp +261
```
if (surfCaps.currentExtent.width == -1)
```
`struct VkExtent2D` is composed of `uint32_t`, so `width` can never be `-1`.

I haven't looked at the code that would set it to that value, but it should probably be written some other way. Maybe add a bool to the struct?
It's also possible the test wasn't meant to be for `-1` but `0`?

1ace:
Make use of them by configuring your build type to debug, for instance
by adding `-DCMAKE_BUILD_TYPE=Debug` to your cmake invocation.

Note that you will get a rather large amount of warnings, but most of those should disappear once you pull #92, #94 & #95, and fix #90.
#99 will also get rid of a warning, but I'm not entirely sure it's the right solution. Your call :)
#100 will also fix one those by virtue of upgrading `glm`: g-truc/glm#495

There will still be a couple unused functions, but I think that's a bug and they should be used, not removed. See #102

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -29,6 +29,9 @@ message(STATUS ${VULKAN_LIB})
 # Set preprocessor defines
 set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -DNOMINMAX -DVK_PROTOTYPES -D_USE_MATH_DEFINES")
 
+# Debug flags
+set(CMAKE_CXX_FLAGS_DEBUG "${CMAKE_CXX_FLAGS_DEBUG} -Wall -Wextra -Wundef")
+
 add_definitions(-D_CRT_SECURE_NO_WARNINGS)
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Got rid of most of the compiler warnings (with above flags) from my base classes and most that are left are from e.g. glm and gli (even though I've updated these). Will take a look at those left eventually.

1ace:
-isystem should help with that :)

On 23 April 2017 11:09:27 BST, Sascha Willems <notifications@github.com> wrote:
>Got rid of most of the compiler warnings (with above flags) from my
>base classes and most that are left are from e.g. glm and gli (even
>though I've updated these). Will take a look at those left eventually.

Cheers,
  Eric

PR #107, opened 2016-03-25
Error on preproccesor for unsupported platforms 

jeffro256:
I added an error directive to catch unsupported platforms. If someone tried to compile this on OSX for example, it would say that Vulkan is not supported instead of being like "App is missing an entry point" or something like this. Its just a small change that points in the way of the actual error if someone were to compile on OSX or other unsupported platform.

### DIFF
--- a/bloom/bloom.cpp
+++ b/bloom/bloom.cpp
@@ -1349,6 +1349,8 @@ void android_main(android_app* state)
 #elif defined(__linux__)
 // Linux entry point
 int main(const int argc, const char *argv[])
+#else
+#error Vulkan not supported by platform
 #endif
 {
 #if defined(__ANDROID__)
--- a/computeparticles/computeparticles.cpp
+++ b/computeparticles/computeparticles.cpp
@@ -784,6 +784,8 @@ void android_main(android_app* state)
 #elif defined(__linux__)
 // Linux entry point
 int main(const int argc, const char *argv[])
+#else
+#error Vulkan not supported by platform
 #endif
 {
 #if defined(__ANDROID__)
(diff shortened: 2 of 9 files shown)


1ace:
Good idea indeed, and the patch looks good to me. :+1:

Three nit-picks though:
- Could you squash those two commits?  There's no reason to split them.  
You should also use the title of this PR as the commit message, as it provides useful information not in the current commit messages.
- `#error` doesn't need quotes, so those are superfluous and will be printed as-is.
- There are unrelated changes of end-of-file character. While I do prefer ending every source file with a newline and so does my editor, I usually avoid committing that as it is unrelated to the real change here. *(I kinda want you too leave them though, as it would mean I wouldn't have to do a partial commit everytime xD)*

Feel free to ignore me though, I'm not even a maintainer for this project :stuck_out_tongue_closed_eyes:

jeffro256:
@1ace Yeah those are are good ideas.
  1. How do you combine 2 commits? I wanted them to be as one, but it didn't work out that way...
  2. I didn't know that `#error` doesn't need quotes. That's nice to know. [https://gcc.gnu.org] had it so I did.
  3. Sorry I guess `nano` automatically removes the newline, because I sure didn't

jeffro256:
DO NOT MERGE!!!!! Something went horribly wrong...

jeffro256:
Okay I think this new one is better

SaschaWillems:
Thx for your pull request. I'm a bit busy right now, but I'll take a look at it. Also note that OSX and iOS support for Vulkan will be available soon with MoltenVK (though that's not native). I haven't seen it yet, but I suspect that they'll add in their own wsi extensions for OSX and iOS.

1ace:
I see you figured it out without my help :+1:

Did you modify the error message?  I don't remember what it was, but `Vulkan is not supported on this platform` would sound better IMO.

SaschaWillems:
Note sure if this is still required with Vulkan now having surface extensions for pretty much every platform (incl. Mac OSX thx to MoltenVK), so closing this for now.

PR #109, opened 2016-03-25
Revert unintended changes to distancefieldfonts demo

nanley:
Commit a09c16950544aff2e55faee580c5cf13ffd8183a, did more than updating
android application labels (its stated purpose). Revert the extra
changes that cause the distancefieldfonts demo to render differently
than the provided screenshot.


SaschaWillems:
Thx. This was a mistake. I'm currently adding Android support to all examples and noticed a possible AMD driver bug in this example, that's where the changes are coming from.

SaschaWillems:
I pushed android support for this example and also slightly changed the blend setup.

PR #112, opened 2016-03-26
Decoupled VulkanExample from VulkanExampleBase 

GoldSciences:
This version of the samples is the first step I'm taking in order to create a multilayered portable framework from your code. This first step is intended to reduce redundancies and to hide most of the portability issues into the new CVulkanFramework class (AKA VulkanExampleBase) and makes use of two new abstract interfaces (IVulkanGame and IVulkanFramework). Currently work in progress of course as this is a huge task, but at least I got to build and run all of the samples without breaking anything (for windows builds at least). One of the most notable improvements is that now the entry point is located in a single place instead of being redundantly repeated (and refactored and upgraded btw) for every example.

One of the problems I'm facing at the moment is that I'm not at home and I'm unable to build and test linux or android versions, so maybe something got broken for those builds in the process of merging. You can use this code or keep yours as it is, though you would help me a lot if you took this addition as part of your samples.


SaschaWillems:
Wow, that's a huge change. While I appreciate all the effort you put into this I'd rather not like to merge this into my examples and leave the relation between the examples and the base class in it's current state, as that's the coding style I've settled for my examples and I'd like to stick with that.

While I understand that decoupling may be desirable, that was not the intention of my examples, though you're right with the multiple entry points, which in turn is not desirable at all.

Maybe [Norbert Nopper's examples](https://github.com/McNopper) are more what you're looking for as he is heavily abstracting using Interfaces.

GoldSciences:
Oh no problem, as I saw you reworking the entry points redundantly I thought it wasn't differing enough from your code yet, but you're right that I'm taking a completely different approach and not far from now they will competely different things. Nice work btw; I hope to find the time to learn enough about vulkan and certainly your examples are helping a lot.

Issue #113, opened 2016-03-27
compilation error


luckykatahanas:
https://github.com/SaschaWillems/Vulkan/blob/master/distancefieldfonts/distancefieldfonts.cpp#L168

I'm not too familiar with filebuf, but in C++98 and C++11 it never took a FILE* in its coinstructor.

http://www.cplusplus.com/reference/fstream/filebuf/open/

/home/stan/working/Vulkan/distancefieldfonts/distancefieldfonts.cpp: In member function ‘void VulkanExample::parsebmFont()’:
/home/stan/working/Vulkan/distancefieldfonts/distancefieldfonts.cpp:168:31: error: no matching function for call to ‘std::basic_filebuf<char>::basic_filebuf(FILE*&)’
   std::filebuf fileBuffer(file);
                               ^
In file included from /usr/include/c++/5.3.0/fstream:1081:0,
                 from /home/stan/working/Vulkan/base/vulkantools.h:17,
                 from /home/stan/working/Vulkan/base/vulkanexamplebase.h:36,
                 from /home/stan/working/Vulkan/distancefieldfonts/distancefieldfonts.cpp:25:
/usr/include/c++/5.3.0/bits/fstream.tcc:94:5: note: candidate: std::basic_filebuf<_CharT, _Traits>::basic_filebuf(std::basic_filebuf<_CharT, _Traits>&&) [with _CharT = char; _Traits = std::char_traits<char>]
     basic_filebuf<_CharT, _Traits>::
     ^
/usr/include/c++/5.3.0/bits/fstream.tcc:94:5: note:   no known conversion for argument 1 from ‘FILE* {aka _IO_FILE*}’ to ‘std::basic_filebuf<char>&&’
/usr/include/c++/5.3.0/bits/fstream.tcc:79:5: note: candidate: std::basic_filebuf<_CharT, _Traits>::basic_filebuf() [with _CharT = char; _Traits = std::char_traits<char>]
     basic_filebuf<_CharT, _Traits>::
     ^
/usr/include/c++/5.3.0/bits/fstream.tcc:79:5: note:   candidate expects 0 arguments, 1 provided

PR #114, opened 2016-03-27
distancefieldfonts: fix SaschaWillems/Vulkan#113

luckykatahanas:


SaschaWillems:
Thx, nice catch :+1: This compiled fine with VS2015 and with clang from the NDK, but I didn't test with MinGW this time, so this slipped through.

PR #119, opened 2016-04-03
Fix compilation of multithreading example on Debian with CMake 3.5.1

pH5:
Compilation of the multithreading example on Debian with CMake 3.5.1 would
fail with the following error because ${PTHREAD} was not set:
    [ 61%] Linking CXX executable bin/multithreading
    /usr/bin/ld: CMakeFiles/multithreading.dir/multithreading/multithreading.cpp.o: undefined reference to symbol 'pthread_create@@GLIBC_2.2.5'
This patch uses ${CMAKE_THREAD_LIB_INIT} [1] instead, initialized using
find_package(Threads REQUIRED).

[1] https://cmake.org/cmake/help/v2.8.12/cmake.html#module:FindThreads


SaschaWillems:
Works fine with Ubuntu too. Thx :+1:

PR #123, opened 2016-04-07
FPS counter is not visible on Linux/XCB and nvidia drivers

zaps166:
Add missing ```xcb_flush()```, see the example: http://www.x.org/archive/X11R7.7/doc/man/man3/xcb_change_property.3.xhtml :)

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -468,6 +468,7 @@ void VulkanExampleBase::renderLoop()
 			xcb_change_property(connection, XCB_PROP_MODE_REPLACE,
 				window, XCB_ATOM_WM_NAME, XCB_ATOM_STRING, 8,
 				windowTitle.size(), windowTitle.c_str());
+			xcb_flush(connection);
 			fpsTimer = 0.0f;
 			frameCounter = 0.0f;
 		}


zaps166:
Hmm, without this flush the window title won't change on nvidia drivers, so the FPS counter is invisible on Linux (AFAIK it works on Intel drivers, but not on nvidia). The XCB documentation also shows that flush must be called after changing the window title. Qt also uses flush after changing the window title: https://github.com/qtproject/qtbase/blob/5.6/src/plugins/platforms/xcb/qxcbwindow.cpp#L1543

SaschaWillems:
Sorry for the late reply. I'm currently writing an example for rendering text as an overlay (using a second render pass) to the existing scene and I plan on adding this functionality to the example base class later on, so that the examples no longer use the window title bar to report performance numbers.

zaps166:
Ok, so I'm waiting for this feature :)

SaschaWillems:
The examples use a text overlay to display a fps counter, so no more need to put this into a window title.

zaps166:
Ok, but anyway it still tries to update a window title, but it doesn't work (wasting CPU time, inconsistency between drivers) :)

SaschaWillems:
Yes, I'm currently doing some refactoring on the example base class and will address that too.

PR #129, opened 2016-04-10
[WIP] Add GPU selection

1ace:
DO NOT MERGE! :stuck_out_tongue_closed_eyes:

This is as far as I got tonight. No tested, and I'm pretty sure it's not done, but I can think anymore.
Just sending it out so you know i'm working on it, and so you can tell me if you want to do things differently :]

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -562,7 +562,7 @@ VkSubmitInfo VulkanExampleBase::prepareSubmitInfo(
 	return submitInfo;
 }
 
-VulkanExampleBase::VulkanExampleBase(bool enableValidation)
+VulkanExampleBase::VulkanExampleBase(bool enableValidation, size_t deviceSelected)
 {
 	// Check for validation command line flag
 #if defined(_WIN32)
@@ -583,7 +583,7 @@ VulkanExampleBase::VulkanExampleBase(bool enableValidation)
 
 #if !defined(__ANDROID__)
 	// Android Vulkan initialization is handled in APP_CMD_INIT_WINDOW event
-	initVulkan(enableValidation);
+	initVulkan(enableValidation, deviceSelected);
 #endif
 
 #if defined(_WIN32)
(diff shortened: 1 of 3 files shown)


SaschaWillems:
Yes, this comes close to what I've thought about (and started working on). I'm not sure about how I'll handle the selection, but having a new command line argument seems to be the easiest way, so I guess I'll go with that one ;)

base/vulkanexamplebase.cpp
```diff
@@ -684,11 +684,52 @@ void VulkanExampleBase::initVulkan(bool enableValidation)
...
+
+#if 1
+		// Use the first physical device reported if no valid value is provided
+		deviceSelected = 0;
+		std::cout << "Using #" << deviceSelected << std::endl;
+		std::cout << "Choose an other device from the list with `--gpu`" << std::endl;
+#else
+		// Require user to choose a device
+		vkTools::exitFatal("Please choose a device from the list with `--gpu`");
+#endif
```
1ace: Forgot to ask: which of these two do you prefer?
On invalid/missing user selection, if there's more than one device:
- select first device
- print an error and abort

I'm leaning towards the former (enabled here), as users are notorious for never reading error messages and would open issues instead of just providing the argument requested.

1ace:
> and started working on

Are far are you? We might as well merge our work (since you said it's the same idea, so it should be compatible).
The question is then: do I merge your changes, or do you merge mine?  ie.: **who finishes the work?**

I guess I should've asked/told you before starting to work on it, to avoid doing the same work twice. :grin:

SaschaWillems:
> Forgot to ask: which of these two do you prefer?
On invalid/missing user selection, if there's more than one device:
select first device
print an error and abort

Select the first device. Device name is displayed in the title bar anyway.

> Are far are you? We might as well merge our work (since you said it's the same idea, so it should be compatible).

Almost finished, but my spare time is very limited right now, so I don't know if I'll be able to tackle this before the weekend.

1ace:
OK, letting you finish then.
Feel free to use whatever code I wrote that might be useful to you. You can close this issue once you're done :]

SaschaWillems:
Better late than never. Just implemented this now that the command line argument handling has been reworked and is working on linux too. The example can now list all available Vulkan capable GPUs (-listgpus) and you can specify the GPU (index) to use via -g (or -gpu).

PR #138, opened 2016-04-21
CMAKE_CXX_FLAGS is overwritten when later adding /Esc option.

d3x0r:


SaschaWillems:
Thx for spotting this :+1:

PR #139, opened 2016-04-24
Fix race condition where 'visible' result is checked before job finishes

baldurk:
This code in multithreading.cpp is broken at the moment.

The problem is that `threadRenderCode()` called inside the job is what updates the `threadData[t].objectData[i].visible` variable, which defaults to true. It's almost certain that the job won't run before the visible variable is checked.

This means the first time through the code the majority of objects will be rendered even if the visibility check ultimately fails and the command buffer isn't recorded, which I don't believe is valid. A more minor problem is that you will submit old command buffers for objects for an extra frame that go out of view from visible to invisible.

There's a couple of ways to fix this - a more complex but perhaps most optimal solution would be to build a pre-thread list of command buffers for visible objects in parallel then do a final coalesce of each thread's results into the final buffer just with `numThreads` copies (or call vkCmdExecuteCommands once per thread list, but that's probably less efficient on the GPU).

I think this PR is the simplest and cleanest solution though, since there's only a small number of objects. I just split up the code to iterate over the objects again to build the list of command buffers after the sync point.

### DIFF
--- a/multithreading/multithreading.cpp
+++ b/multithreading/multithreading.cpp
@@ -432,21 +432,28 @@ public:
 		updateSecondaryCommandBuffer(inheritanceInfo);
 		commandBuffers.push_back(secondaryCommandBuffer);
 
+		// Add a job to the thread's queue for each object to be rendered
 		for (uint32_t t = 0; t < numThreads; t++)
 		{
-			// Add a job to the thread's queue for each object to be rendered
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Thanks for this PR :+1: 

I haven't had the time to dig deeper into this, but will try to merge this soon.

PR #146, opened 2016-05-08
Add use of VK_EXT_debug_marker

baldurk:
Hey! I'm playing evangelist for a bit :).

This PR adds the use of new functions exposed in VK_EXT_debug_marker, to allow object labelling and command buffer annotation. Currently RenderDoc supports this in the latest version (v0.29), but I imagine other tools and likely the validation layers will support it in future.

I've committed the latest header from the KhronosGroup/Vulkan-Docs repository, but if you want you can wait until the next SDK drop to merge this and replace that commit with the vulkan header from that SDK, it's up to you.

I also added example usage to the bloom sample - you don't have to take that commit either I just wanted to show what it would look like. Here's some screenshots of RenderDoc showing some of what it does:

Buffer naming

Shader and image naming

Command buffer annotation

### DIFF
--- a/base/vulkandebug.cpp
+++ b/base/vulkandebug.cpp
@@ -88,4 +88,138 @@ namespace vkDebug
 			// DestroyDebugReportCallback(instance, msgCallback, nullptr);
 		}
 	}
+
+	PFN_vkDebugMarkerSetObjectNameEXT DebugMarkerSetObjectName = NULL;
+	PFN_vkCmdDebugMarkerBeginEXT CmdDebugMarkerBegin = NULL;
+	PFN_vkCmdDebugMarkerEndEXT CmdDebugMarkerEnd = NULL;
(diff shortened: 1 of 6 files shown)


SaschaWillems:
Awesome :smile: 

Just downloaded the recent RenderDoc release and testing with a more advanced Vulkan demo and VK_EXT_debug_marker sounded like an awesome idea. 

Will add this to all of the examples this week :+1:

SaschaWillems:
Finally got around merging this. Note that I did not merge the changes to the bloom example but only the changes to the base files.

This is because rather than just adding the debug markers to an existing example I'll be doing a separate example to demonstrate the debug marker extension and also want to do a small write up (or tutorial) on implementing the debug marker and using it RenderDoc.

baldurk:
Thanks! That's totally fine, I just included the bloom example to show how it could be used :smile:.

PR #147, opened 2016-05-10
Fix asset filename in multisampling

TheBB:
Looks like voyager.obj went missing. (The .dae file is used in the mesh example.)


SaschaWillems:
Thx :+1:

PR #151, opened 2016-05-16
Adjust occlusionquery according to Vulkan spec.

pbantolas:
This moves vkCmdCopyQueryPoolResults outside of the Render Pass, as it
should be according to the Vulkan spec. This understandably looks
wrong for the first frame, but by doing this you avoid having to use two
render passes to get the occlusion result.

### DIFF
--- a/occlusionquery/occlusionquery.cpp
+++ b/occlusionquery/occlusionquery.cpp
@@ -261,17 +261,6 @@ public:
 
 			vkCmdEndQuery(drawCmdBuffers[i], queryPool, 1);
 
-			// Query results
-			vkCmdCopyQueryPoolResults(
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Thx for noticing :1 Taking a look at this I also noticed that I do a copy and a get of the results which is unnecessary too.

pbantolas:
Good points, just saw that. So the CopyQueryPoolResults and it's accompanying buffer is not needed at all then.

SaschaWillems:
This also fixed the example on my Ivybrdige-based Intel Ultrabook on Linux using the Mesa drivers. One more demo running on min. spec hardware :+1:

PR #158, opened 2016-05-23
Fix clear value array in textoverlay

nanley:
According to section 7.4. Render Pass Commands of the Vulkan spec,
VkRenderPassBeginInfo::pClearValues is indexed by attachment number.
Since the second attachment in the renderpass will be cleared, a clear
value array of length two is required, and the second element of that
array must be assigned the desired clear value.

### DIFF
--- a/textoverlay/textoverlay.cpp
+++ b/textoverlay/textoverlay.cpp
@@ -632,14 +632,14 @@ public:
 	{
 		VkCommandBufferBeginInfo cmdBufInfo = vkTools::initializers::commandBufferBeginInfo();
 
-		VkClearValue clearValues[1];
-		clearValues[0].color = { { 0.0f, 0.0f, 0.0f, 0.0f } };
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Thx for your PR. With commit https://github.com/SaschaWillems/Vulkan/commit/19e108165c272c73fcfc9008de99c44cc253142b#diff-b777f7a9f0fb820e21a983ed7fed5bd3 I already changed the loadOp for the depth attachment to DONT_CLEAR so this should not be necessary anymore.

nanley:
The commit you're referring to seems to be for the text overlay in that appears in several of the Vulkan demos. I am not seeing any issue with that. This PR is meant to fix the textoverlay example that currently crashes on the Intel Vulkan driver.

SaschaWillems:
Sorry my fault. Will take a look at this.

nanley:
No problem. Thanks for the merge!

Issue #156, opened 2016-05-18
You are not allowed to transition layouts when the image is not acquired.


ratchetfreak:
https://www.khronos.org/registry/vulkan/specs/1.0-wsi_extensions/xhtml/vkspec.html#_wsi_swapchain

> Use of a presentable image must occur only after the image is returned by vkAcquireNextImageKHR, and before it is presented by vkQueuePresentKHR. **This includes transitioning the image layout and rendering commands.**

https://github.com/SaschaWillems/Vulkan/blob/master/base/vulkanswapchain.hpp#L374

This will end up doing an image transition on a swapchain image that is not owned by the application. Instead you can always transition from UNDEFINED in the pre-present transition.

SaschaWillems:
Thx for pointing me at that part of the spec. When I wrote the swapchain class the only example code available was the one from the LunarG examples and they also do this.

But I wonder why validation doesn't catch this (and why it works on all GPUs I have tested so far).

If such an image layout transition is not allowed for images not owned by the application (which makes sense) then there should be a validation that catches this.

ratchetfreak:
The validation isn't perfect yet [and there is an open issue about it](https://github.com/KhronosGroup/Vulkan-LoaderAndValidationLayers/issues/412).

SaschaWillems:
Thx for that link. Will fix this.

PR #169, opened 2016-05-30
Don't perform layout transitions on unacquired swapchain images

jherico:
Fix for #156


SaschaWillems:
Thx :+1: Will check on different implementations, and if it works as expected will merge.

SaschaWillems:
The examples no longer use pre- and post present barriers and do layout transitions via subpasses, so closing this one.

PR #170, opened 2016-05-30
Tweaking build and cmake stuff

jherico:
* Fixed a number of examples still used key handling in a custom WndProc function instead of using the `keyPressed` method
* Moved all boilerplate code for executing each example (about 60 lines per file) to macros
* Modified the visual studio 2015 projects to consolidate the shared code into a static library project wich all other examples now reference.  
* Consolidated all example code header boilerplate into "vulkanexamplebase.h"
* Update the cmake configuration to work out of the box for Visual Studio 2015, using a static library for common code
* Enable validation by default in debug builds

Moving the existing common cpp files out of each individual example and into a shared library base project significantly reduces the build time for me.  It could potentially be reduced even further by having the precompiled header generated by the base project shared by all the other projects, but that's a little hacky, so I haven't done it.


SaschaWillems:
Thanks for all the effort you have put into this pull request, but it contains to much stuff to simply merge it and also conflicts with some of the changes I'm currently making, so please don't be offended if I'm not going to merge it, especially as I want to make sure that everything compiles out of the box for windows, linux and android using different compilers and tool chains.

There are several issues with this:

Key handling contains windows specific key codes, this will fail on Linux and Android. I'm adding these to the examples currently for multiple platforms (along with info in the text overlay).

Enabling validation by default for debug builds arguable. I wouldn't want it, as I'm always using layers compiled from current sources with the debug config, so enabling them will result in a huge performance hit. This would also fail running examples in debug config if the SDK (and layers) are not installed.

Moving the example construction boiler plate to a macro is on my todo list and something I'll be taking a look at in the future, same for the library thing for common base functionality.

jherico:
I made the commits pretty targeted for this, so I'll see if I can break it up into several individual PRs that target specific things that aren't contentious.

PR #183, opened 2016-06-16
Fix a warning

xyproto:
`MAX_CHAR_COUNT` exists in two places and resulted in a warning when compiling the project.
- gcc 6.1.1
- cmake 3.5.2
- ninja 1.7.1
- 64-bit Arch Linux


SaschaWillems:
Fixed with https://github.com/SaschaWillems/Vulkan/commit/ff643e5e48336cc2d0b819f45dbc705a82d2fb67

PR #186, opened 2016-06-18
Make the keys work on Linux

xyproto:
* Added defines for keyboard scancodes on Linux in `vulkanexamplebase.h`.
* Moved the hardcoded values for keys on Windows into `vulkanexamplebase.h` as well, so that for instance `KEY_S` will work regardless of platform.
* Minor changes to the formatting (removed trailing whitespace a couple of places).

All the keybindings of all the Vulkan samples now work here, on Linux.

It might be even better to use an existing library for handling keys, like SDL2, but this pull request should solve the problem for now.

### DIFF
--- a/base/vulkanexamplebase.h
+++ b/base/vulkanexamplebase.h
@@ -50,9 +50,59 @@
 #define GAMEPAD_BUTTON_R1 0x1005
 #define GAMEPAD_BUTTON_START 0x1006
 
+#if defined(__linux__) && !defined(__ANDROID__)
+
+#define KEY_W 25
+#define KEY_A 38
+#define KEY_S 39
+#define KEY_D 40
+
+#define KEY_Q 24
+#define KEY_E 26
+#define KEY_F 41
+#define KEY_B 56
+#define KEY_O 32
+#define KEY_N 57
+#define KEY_T 28
+#define KEY_L 46
+
+#define KEY_LEFT 113
+#define KEY_RIGHT 114
+#define KEY_UP 111
+#define KEY_DOWN 116
+
+#define KEY_SPACE 65
+#define KEY_RETURN 36
+#define KEY_ESC 9
+
+#define KEY_LSHIFT 50
+#define KEY_RSHIFT 62
+
+#define KEY_NP_PLUS 86
+#define KEY_NP_MINUS 82
+
+#else
+
+#define KEY_W 0x57
+#define KEY_S 0x53
+#define KEY_D 0x44
+
+#define KEY_B 0x42
+#define KEY_O 0x4F
+#define KEY_N 0x4E
+#define KEY_T 0x54
+#define KEY_L 0x4C
+
+#define KEY_SPACE 0x20
+
+#define KEY_NP_PLUS 0x6B
+#define KEY_NP_MINUS 0x6D
+
+#endif
+
 class VulkanExampleBase
 {
-private:	
+private:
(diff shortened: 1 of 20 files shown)


SaschaWillems:
First, thx for putting in all that effort to make this work :+1: 

Can't promise if I'll merge this request as it's affecting lots of files, a few of I'm currently working on. So in the future it may be a good idea to ask in advance :wink:

base/vulkanexamplebase.h
```diff
@@ -50,9 +50,59 @@
 #define GAMEPAD_BUTTON_R1 0x1005
 #define GAMEPAD_BUTTON_START 0x1006
 
+#if defined(__linux__) && !defined(__ANDROID__)
```
1ace: I think I'd prefer Android and Linux to be separate branches, even if one is currently empty. E.g.:
```
#if defined(__ANDROID__)
#elif defined(__linux__)
#define ...
```
This makes the behaviour of an `#else` branch more predictable *(currently, Android, Windows and anything else fall in that `#else` on line 84)*, and makes it easier to add Android-specific code later.

base/vulkanexamplebase.h
```diff
@@ -50,9 +50,59 @@
...
+#define KEY_RETURN 36
+#define KEY_ESC 9
+
+#define KEY_LSHIFT 50
+#define KEY_RSHIFT 62
+
+#define KEY_NP_PLUS 86
+#define KEY_NP_MINUS 82
+
+#else
```
1ace: I think it would be better to explicitly say what those codes are for.
If I understand correctly, the `#else` branch here if for Windows, so `#elif defined(_WIN32)` would be better IMO.

base/vulkanexamplebase.h
1ace: Speaking of: aren't those `GAMEPAD_BUTTON_*` Android-specific?

xyproto: The `#ifdef` checks follows the pattern that is already there in other files.

The `GAMEPAD_BUTTON_*` related code might very well be Android-specific, but I did not touch that code as part of this patch.

base/vulkanexamplebase.h
xyproto: If doing checks with `defined(_WIN32)` instead, several other source files would need to be changed as well. This is likely to be a good idea, but I feel like this belongs in a different pull request, and is a different thing.

1ace: Agreed, consistency is a good argument, although I feel like progressive enhancement could also be good. You're welcome to make that other PR then :)

xyproto: The check I used in the pull request works well and is in line with the checks that is already used in the project. It's usually considered a good thing that pull requests try to follow the existing code style, not a disadvantage.

For this pull request, I disagree that using `defined(_WIN32)` would be better, since it would be different from the existing code style that is already in place.

jherico:
I've done something similar, but more extensive in a [commit](https://github.com/jherico/Vulkan/commit/6f4ff6b4c839b03a2a7616981f0105b26be43a00) on my fork.  It completely replaces all of platform window handling on Linux and Win32 with GLFW, and uses the `GLFW_KEY_*` defines.  For android, I've explicitly added a set of defines for the `GLFW_KEY_*` macros that are currently used.

SaschaWillems:
Fixed using a separate header for keycodes depending on platform and the PR from #219.

PR #191, opened 2016-06-27
Use CMake more effectively

jherico:
This PR is modifies the existing CMake file to do a number of things
- Make all dependencies into external projects which are fetched on demand
- Create a shared project for the common code
- Improve project and solution organization
  - Shader source files referenced by a project are now listed directly in the project itself
  - Eventually I'd like to fix the CMake to detect changes to the shaders and recompile them on demand

Additionally, all of the OBJ files have been modified to remove extraneous whitespace on the vertex definition lines, because this was throwing off the version of Assimp I'm using.


jherico:
Note, I have a [subsequent commit](https://github.com/jherico/Vulkan/commit/6f4ff6b4c839b03a2a7616981f0105b26be43a00) that moves all of the Linux and Window platform specific code to use GLFW instead.  I can include that in this PR, or I can include a distinct PR with those changes later.

PR #198, opened 2016-07-10
Fix getMemoryType not checking actual number of memory types

Overv:
The function is currently iterating over the maximum number of possible memory types (which should be `VK_MAX_MEMORY_TYPES` instead of `32`) instead of the actual number.

### DIFF
--- a/base/vulkanTextureLoader.hpp
+++ b/base/vulkanTextureLoader.hpp
@@ -44,7 +44,7 @@ namespace vkTools
 		// Get appropriate memory type index for a memory allocation
 		uint32_t getMemoryType(uint32_t typeBits, VkFlags properties)
 		{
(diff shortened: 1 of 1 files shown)


SaschaWillems:
While VK_MAX_MEMORY_TYPES is defined as 32, using the define instead of a fixed number is the correct way. Thx for hinting me at this, will change this later today.

SaschaWillems:
As part of a larger refactoring (VulkanExampleBase has become too much of a singleton) I have moved the memory type index selection to the new [VulkanDevice](https://github.com/SaschaWillems/Vulkan/blob/master/base/vulkandevice.hpp) unit that checks against number of actual memory types reported. I also replaced all the memory type index functions scattered across the samples, so now only this one will be used.

PR #201, opened 2016-07-11
Linux: Fix name of embedded libvulkan.so

akien-mga:
Fixes this issue when running binaries that look for the _soname_, i.e. `libvulkan.so.1`:
```
./bin/bloom
./bin/bloom: error while loading shared libraries: libvulkan.so.1: cannot open shared object file: No such file or directory
```

Also made the renamed `libvulkan.so.1` executable (0755) and added the expected `libvulkan.so` symbolic link.


1ace:
(Tip for next time: use `git mv a b` instead of `mv a b; git add .` so that git sees that as a rename, not a delete+create)

akien-mga:
Hm, I'm pretty sure I did use `git mv`; I had then to `git add` the change again due to `chmod +x` on the moved file, maybe that broke it?

1ace:
It doesn't really matter, don't change your PR just for that :)
Also, while some distributions still add the executable bit, its historical use hasn't applied for decades. These distributions add it just because "it doesn't hurt to have it". *(Again, no need to change your PR :))*

akien-mga:
Actually my distribution ensure that libraries have the executable bit at least so that RPM can extract debug info from it. I'm not sure if it has another purpose nowadays, but since the compiler (at least gcc) generates .so files as 0755, I guess it doesn't hurt to keep it this way for consistency.

SaschaWillems:
@akien-mga: Thx for your PR :+1:. I'm currently a bit short of time so it may take me a few days or weeks before I can pull this into master.

PR #220, opened 2016-08-14
queue submitted twice

heeen:
It looks like the queue is submitted twice and synchronized once with wait idle and once with a fence.


heeen:
If this is on purpose, could you explain the reasoning?

SaschaWillems:
No, that was not on purpose. The submit info you removed was a left-over. Thx for the PR :+1:

PR #221, opened 2016-08-14
Unrecognized keyboard input in mesh example on Linux

pluehne:
This fixes a minor issue on Linux systems, where the Windows-specific key code is not working for toggling the visibility of the wireframe. Instead, I used the `KEY_W` macro defined for multi-platform purposes.


SaschaWillems:
Thx

PR #222, opened 2016-08-15
Missing text overlay in tesselation example

pluehne:
This is a very minor issue. By accident, I found out that the key `s` toggles the split screen. As this is currently not indicated, I added a corresponding text overlay.


SaschaWillems:
Yes, I think there are a few other examples where some of the keys are not displayed in the text overlay. Will have to check that.

Thx for the PR.

PR #233, opened 2016-09-17
cmake: Add scenerendering demo to CMakeLists.txt

aunali1:
Simple commit that adds the scenerendering demo back into CMakeLists.txt so that it builds on Linux and macOS.


SaschaWillems:
Thx

PR #236, opened 2016-09-21
Assigned value to a "Normal" variable in ground.vert file.

Ekzuzy:
Fixed bug with a variable consumed by fragment shader not being assigned in a vertex shader. Updated spv file. Added bat file for automatic SPIR-V generation.


SaschaWillems:
Thx 👍 Oddly my validation layers (built from source) won't complain about this, even though they should.

Just a quick note for the .bat files for shader compilation: I added a python script some time ago in the shader root folder that can be used to generate SPIR-V binaries for all shaders in a given directory. That's why I don't add .bat files to new projects anymore.

Issue #168, opened 2016-05-30
Why does multithreading example use per-thread vertex and index buffers?


jherico:
I noticed this when I saw validation errors on the example.  The mesh loader does not apply give the buffer memory the TransferSrc flag, so the copy command in setting up the per-thread buffers generates the errors.  However, I'm not clear on why each thread needs it's own copy of the static mesh data.  Replacing the copy code in `prepareMultiThreadedRenderer` with `thread->mesh = meshes.ufo` seems to work just fine and produce no validation errors.

SaschaWillems:
This dates back to when the buffers weren't device local and may not be necessary anymore. I'll take a look at it this weekend.

As for the validation error from the missing usage flags, see #164.

jherico:
ok, I was just trying to understand if I was breaking something. thanks.

dmarian:
For the record, I adapted this example. I use a single vertex and index buffer for all threads and on each thread I do:

vkBeginCommandBuffer(...);
vkCmdSetViewport(...);
vkCmdSetScissor(...);
vkCmdBindPipeline(...);
vkCmdBindVertexBuffers(...);
vkCmdBindIndexBuffer(...);
for_each (object) {
  vkCmdPushConstants(...);
  vkCmdDrawIndexed(...);
}
vkEndCommandBuffer(...);

SaschaWillems:
Yes, I tested that too some weeks ago and it works fine. Haven't found the time to update this example yet though.

PR #237, opened 2016-09-21
Added missing usages in buffers used as a source of data copy.

Ekzuzy:
Added missing usages in buffers used as a source of data copy. Fixed problems with types in starsphere.frag shader. Updated starsphere.frag.spv file.


SaschaWillems:
Thx for the PR. As for the buffer flags, see issue #168. I have just pushed a commit that removes the buffer copies for the MT example, so the buffer flags for transfer src are no longer required, so no more need for that.

But the shader changes make sense, so I'll port them over 👍

PR #238, opened 2016-09-21
Fixed problems with layout transitions and color formats.

Ekzuzy:
Fixed problems with layout transitions in deferred and deferredshadows samples. Fixed mismatched color formats of attachments/render passes/framebuffers. Removed unused variables.

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -227,7 +227,7 @@ void VulkanExampleBase::prepare()
 			vulkanDevice,
 			queue,
 			frameBuffers,
-			colorformat,
+			swapChain.colorFormat,
 			depthFormat,
 			&width,
 			&height,
@@ -1520,7 +1520,7 @@ void VulkanExampleBase::setupRenderPass()
 {
 	std::array<VkAttachmentDescription, 2> attachments = {};
 	// Color attachment
-	attachments[0].format = colorformat;
+	attachments[0].format = swapChain.colorFormat;
 	attachments[0].samples = VK_SAMPLE_COUNT_1_BIT;
 	attachments[0].loadOp = VK_ATTACHMENT_LOAD_OP_CLEAR;
 	attachments[0].storeOp = VK_ATTACHMENT_STORE_OP_STORE;
--- a/base/vulkanexamplebase.h
+++ b/base/vulkanexamplebase.h
@@ -93,8 +93,6 @@ protected:
 	vk::VulkanDevice *vulkanDevice;
 	// Handle to the device graphics queue that command buffers are submitted to
 	VkQueue queue;
-	// Color buffer format
-	VkFormat colorformat = VK_FORMAT_B8G8R8A8_UNORM;
 	// Depth buffer format
 	// Depth format is selected during Vulkan initialization
 	VkFormat depthFormat;
--- a/base/vulkanframebuffer.hpp
+++ b/base/vulkanframebuffer.hpp
@@ -139,7 +139,6 @@ namespace vk
 			attachment.format = createinfo.format;
 
 			VkImageAspectFlags aspectMask = VK_FLAGS_NONE;
-			VkImageLayout imageLayout;
 
 			// Select aspect mask and layout depending on usage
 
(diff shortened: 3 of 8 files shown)


Ekzuzy:
Hi. Any update on this pull request? Did You have time to test it? Without these modifications deferred* samples don't work on Intel graphics.

SaschaWillems:
Sry, totally forgot about this PR. I didn't realize that this is required on Intel graphics, as all the examples run fine on my Haswell Ultrabook running Ubuntu with the MESA drivers.

While most changes look fine, the changes to the depth/stencil final layout (https://github.com/SaschaWillems/Vulkan/pull/238/files#diff-bb472eedb08baf1328e3fe24fbe53158R218) causes validation layer errors for the layered depth map.

Is that change really required to get these running on Intel graphics (I guess this affects windows only?).

Ekzuzy:
I will check it with the newest version of validation layers and try to
figure out why they are compaining. In general. these changes are required,
because the layout images are transitioned to is different than the one
specified inside render pass. And proper transitions are necessary in case
of Intel graphics (but the MESA part You mentioned about is interesting).

2016-11-04 11:04 GMT+01:00 Sascha Willems <notifications@github.com>:

Ekzuzy:
Hello.

I've analyzed the deferredshadows sample once again. I'm wondering if the error You mentioned is reported correctly.
In the deferredshadows You have 3 render passes: 1st for depth data, 2nd for color and 3rd for final deferred pass. In the 1st pass depth/stencil texture needs to be in DEPTH_STENCIL_ATTACHMENT_OPTIMAL layout if we want to render into the texture. But in the 3rd render pass the same image/texture is sampled, fragment shader reads from it. So it can't have DEPTH_STENCIL_ATTACHMENT_OPTIMAL layout. For reading/sampling we should use DEPTH_STENCIL_READ_ONLY_OPTIMAL layout (or SHADER_READ_ONLY_OPTIMAL layout). And the final layout of the 1st render pass is the only reasonable place in which this layout transition should occur.

What do You think?

Ekzuzy:
Any news? As I have written, I think validation layers are wrong here. Spec clearly states in descriptor updates section about layouts: "imageLayout is the layout that the image will be in at the time this descriptor is accessed". And, unfortunately, in the deferred samples, other layout is specified in the descriptor set updates and images are transitioned to other layouts. It does work on Nvidia, but it is a violation of the spec, nevertheless.

SaschaWillems:
I'm currently checking this on AMD hardware to ensure that the changes won't break anything and will be going to check with a recent build of the validation layers to see if the warnings persist.

SaschaWillems:
Just checked on an AMD GCN CPU and I think you're right with the validation layers being wrong. With the current code the shadow attachment is garbled on AMD (but no validation errors), with your changes (that are obviously correct) the shadows render fine but I get validation layer errors, so this may really be an error in the validation. I'll take a deeper look at this (hopefully sometime this week) and see how to report it.

Ekzuzy:
Good news - that I (probably) didn't make any mistake in the PR, and that You agree with my opinion, and that Your code works as expected on the AMD hardware with my modifications. ;-)
Bad news that validation layers are wrong here. This is very confusing for everyone that wants to jump into Vulkan API... :-/
Nevertheless, thanks for the info. I'm waiting for more news from You.

SaschaWillems:
Just a quick note that I'm still actively looking into this and as of now I'm no longer sure if it's really a validation problem. Doing some manual image layout transitions fixes the validation messages, so I think the render pass setup of the example may be wrong.

Ekzuzy:
I think there really may be a problem with validation layers. When the above changes are applied to the code, validation layers report problems only with a layered depth image (the one storing depth values for each light source). There are no problems with layout transitions of other images reported by validation layers.
I've reported an issue on the LoaderAndValidationLayers Github page, maybe they will fix it (and the problem with image layouts in descriptor sets not being tracked).

SaschaWillems:
Thank you very much for reporting this in such detail at the LoaderAndValidationLayers  repo 👍

SaschaWillems:
@Ekzuzy Just checked with the recent valdiation layer fixes and with your changes applied, the messages are gone. Do you want to submit another PR, or is it okay if I apply your changes (the ones that still haven't made it into master) to my current code?

Ekzuzy:
Thanks! :-) I have a newer version of code changes, so You don't have to resolve the conflict. It also fixes problems with SSAO sample. I will prepare a new PR tomorrow at work.

Ekzuzy:
Closing - new PR request will be prepared.

PR #239, opened 2016-10-05
Added FPS values to show on console

MuraliSridhar:
Hi Sascha Willems,

To get the FPS on terminal , I made following changes,
Please look if you can have on your repo.

Thanks,
Murali


SaschaWillems:
While this works I don't see what's the value of this. FPS values are already displayed. Can you explain why you would need this?

citrix123:
I Guess for saving / Extract FPS this is good. Also it can help the API Developers to do regression as well.
We can optimize it more, Some thing like Benchmarks . Eg: Glmark2 , unigine etc

Thanks,
Rahul

SaschaWillems:
See #269, I'll be adding performance metrics and benchmark mode with a result file output in a future update.

citrix123:
Looks Cool !!

PR #242, opened 2016-10-28
Use GLFW for cross platform window management

jherico:
## Changes
- Updates required CMake version to 3.7 for the built-in Vulkan find module
- Integrates GLFW 3.2 
  - as an external project on Windows
  - as a required package on Linux
- Uses GLFW for keyboard and mouse handling on Windows and Linux
- Switches to CMake for Visual Studio project creation and deletes embedded VS project files
- Pushes `GLM_FORCE_RADIANS` and `GLM_FORCE_DEPTH_ZERO_TO_ONE` defines into CMake config
## Current issues
- ~~Untested on Linux~~ Did a successful build on Ubuntu 16.10 with mesa-vulkan-dev
- Untested on Android
- Joystick input currently not implemented
- ~~Does not use `glfwGetRequiredExtensions()`~~
- There should be at least one example showing native window / swap chain setup
- Assimp, GLM, and GLI should be made into external projects / required local packages


vinjn:
One exception, Android is unsupported by glfw.

jherico:
@vinjn Much of the existing code is along the lines of 
```
#if WIN32
#elif ANDROID
#elif LINUX
#endif
```

This PR is attempting to consolidate window and input handling only for Linux and Windows, so these blocks would change to 

```
#if ANDROID
#else
#endif
```

The android paths should remain unaffected, but it should simplify the logic for non-Android builds.

SaschaWillems:
As hinted in other PRs I don't have any plans to use GLFW for now.

PR #243, opened 2016-11-01
Added CMAKE_SOURCE_DIR to runtime output directory

MattGuerrette:
This is a small change for the build. The examples rely on the data directory being relative to the bin/ folder. But this is not the case when the project is built using an editor such as CLion, which requires a more explicit path otherwise it creates the output folder in its own temp sandbox directory.

This will not change how the build has been working, it will only make it work with CLion out of the box.


SaschaWillems:
Thx 👍

PR #248, opened 2016-11-07
Cleanup changes to unify entry points and remove duplicated functions in triangle.

shi-yan:
Hi Sascha,

These two commits are code cleanup changes.

The ways entry points are defined are inconsistent in the code. Some tests are defining their own entry point, whereas others are using the entry point macro. The first commit unifies the way entry points are created by only using the macro.

The second commit removes some duplicated functions from the triangle example which are also defined in its base class vulkanexamplebase.

Thanks,

Shi


SaschaWillems:
Thanks again for your effort 👍 

But can you please revert the changes to the triangle example? It's as verbose as possible in purpose, especially as it contains comments not present in the functions from the base class. If that change is reverted I can happily merge this 😄

SaschaWillems:
All examples (except for the triangle one) now use the unified entry points from VulkanExampleBase.h.

PR #249, opened 2016-11-22
Set VkExampleBase::colorformat after swapChainInit

eoma:
The validation layers gives me errors (and the intel anv driver
crashes) if the framebuffer attachments and render pass attachments
doesn't match.

Only tested on Ubuntu 16.10 with mesa-git. Recent updates to the driver
enables SRGB and UNORM, in that order.


eoma:
I discovered in hindsight that this is related to #238, and can in some parts be considered a duplicate.

eoma:
Closing

PR #250, opened 2016-11-27
Don't tell the GPU to store attachments we don't care about

jherico:
The subpass example currently sets all the attachments to `VK_ATTACHMENT_STORE_OP_STORE`.  However, the only attachment whose contents we care about is the color buffer, which holds the output, so the other attachments should be set to  `VK_ATTACHMENT_STORE_OP_DONT_CARE`.


SaschaWillems:
That's right. Storing attachments not used after the renderpass doesn't make sense and may influence performance. Thx for the PR 👍

PR #264, opened 2017-01-08
allow bundled Vulkan lib also with newer cmakes

lamikr:
Remove REQUIRED flag for the Vulkan library
search that are used if CMake version > 3.7.

This will allow the configuration and build fall back
of searching the internal vulkan library versions that
are bundled within this repo. (In case that the user
has not installed vulkan libraries under /usr)

Signed-off-by: Mika Laitio <lamikr@pilppa.org>


lamikr:
Hi
I encountered the problem that this patch fix after cmake version was updated from 3.5 to 3.71 on Mageia 6 Linux distro that I am using.
After applying this small fix, the configure will not anymore stop if the libvulkan.so is not found from /usr/lib and will instead just printout the warning and then try to find and use the version found from this repos libs/Vulkan dir instead just like it does with cmake versions < 3.7

SaschaWillems:
You're right, Failing when reverting to the bundled version makes no sense. 

Thx for the PR 👍

PR #275, opened 2017-01-31
explicitely choose VK_FORMAT_B8G8R8A8_UNORM

ChristophHaag:
The radv driver renders the examples much brighter than amdgpu-pro.

This has been a known issue for a while that appeared in The Talos Principle too:
https://lists.freedesktop.org/archives/mesa-dev/2016-October/132435.html

As the mailing list post says, radv puts VK_FORMAT_B8G8R8A8_SRGB first in the list to intentionally break applications that just use the first available format. Presumably the Vulkan spec allows drivers to put them into any order, so if applications require a specific one to render correctly, they should always look for it in this list.

Not sure if drivers are allowed not to support `VK_FORMAT_B8G8R8A8_UNORM`, so I put in a fallback to use `surfaceFormats[0]` then.

Screenshot showing the difference:
radv-amdgpu-pro

### DIFF
--- a/base/vulkanswapchain.hpp
+++ b/base/vulkanswapchain.hpp
@@ -225,16 +225,33 @@ public:
 		if ((formatCount == 1) && (surfaceFormats[0].format == VK_FORMAT_UNDEFINED))
 		{
 			colorFormat = VK_FORMAT_B8G8R8A8_UNORM;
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Thx for the PR. Intel Mesa on  Linux has the same problem. Source changes look fine too 👍

PR #276, opened 2017-02-01
Add Wayland WSI support

tomek-brcm:
I've recently added Wayland support, if you're interested. Most of the code is pretty straight forward. There are 2 places you might want to have a closer look.

Key handling in bloom example required a special hack in order to build with Wayland. 

Also I didn't want to change existing WSI selection mechainsm so Wayland selection is a bit weird. You have to de-select direct display and select Wayland. I guess a list of WSI-s would be better.


SaschaWillems:
Thx for this great PR! Will try to merge this weekend. As for WSI selection that's something I'd like to address in the future now that there are a number of platforms available to choose from.

tomek-brcm:
You're welcome. 

I've just cleaned up remaining debug print statements, that managed to sneak into the commit. The updated version is in now.

PR #285, opened 2017-02-15
Use  VK_NULL_HANDLE for null shader module.

tomek-brcm:
The VkShaderModule is defined in vulkan.h as:
VK_DEFINE_NON_DISPATCHABLE_HANDLE(VkShaderModule)


SaschaWillems:
Thx 👍

PR #287, opened 2017-02-20
Enabled warnings as errors for Visual Studio builds.

GoldSciences:
+ Enabled warnings as errors for VS builds.
+ Largely reduced build times by removing redundancies for VS builds.
* Fixed all the incorrect type casts reported by the compiler.
* Fixed uninitialized pointer variables.


GoldSciences:
I believe I forgot to set up the dependencies for all of the projects. I'll fix that in a moment.

GoldSciences:
I can revert the changes in the triangle example if you like. I don't see the point on building over and over the vulkanexamplebase files for the other projects and I believe most people dealing with this code probably knows about linking a simple lib. In my case I did this because my computer is too slow and building 50 times the same files for no reason is something I just can't afford.

SaschaWillems:
Thx for your effort. Moving the example base to a library has been requested several times, but I actually prefer it the way it currently is, so if there are no other advantages rather than I'd prefer not to change that.

I also don't want warnings to be treated as errors. While I do that at work for production stuff I prefer a bit more freedom for my personal projects.

But befor discussing this any further: Does this also affect Linux and Android builds? And if so did you test on Android too? CI is only applied for Windows and Linux currently.

GoldSciences:
This affects only Windows builds because the only files changed are the .sln, the .vcxproj and the .filters files, so if you use other build pipeline than VS it should be the same as before except for a few variables to which I added the initialization but that shouldn't be a problem for any of the builds I suppose (I just added = VK_NULL_HANDLE; in front of some member variables). The warnings I had were mostly casts from size_t to uint32_t or signed/unsigned mismatch but I fixed them all and I believe they shouldn't be any problem for any of the builds.

SaschaWillems:
Okay, if it's windows only I may think about it again. As this is a pretty extensive PR please give me some time to review and decide on it. And thx for all the effort 👍

GoldSciences:
Thanks, it's really helpful to reduce the timing because my notebook is slow and it eats up a lot of battery when building. Sadly I don't own an Android device with Vulkan support to test it but the code is mainly untouched.

There is one of the warnings reported by the static analyzer is that in `VulkanDevice::copyBuffer()` line 441 there is the following check: `assert(src->buffer && src->buffer);` which is redundant, so I thought maybe the intention was to check for `dst->buffer` in one side? If so, I can fix it in my next commit, which would leave all the warnings reported by the static analysis also fixed. 

The warnings I've fixed from the static analysis were just a few:
- A hidden variable due to using the same name in an inner scope (I can't remember the name now).
- The `WinMain` missing the SAL annotations.
- Using uninitialized `shaderStages` variable.
- The one I mentioned about the redundant `src->buffer` assert (pending).

GoldSciences:
Actually, there's one more thing. I've been fooling around with the warning level last night and noticed than when increasing it from 3 to 4 the compiler throws more warnings, some of which being actual bugs that go unnoticed. The most common problems seems to be cause of the naming convention and it's that some parameter variables end up being assigned to themselves instead of copying the value to the member variables. Also, some of the initializers are failing to assign the "flags" parameter to the Vulkan structs.

SaschaWillems:
> The most common problems seems to be cause of the naming convention and it's that some parameter variables end up being assigned to themselves instead of copying the value to the member variables. Also, some of the initializers are failing to assign the "flags" parameter to the Vulkan structs.

Thx for the info. I'll take a look at these.

GoldSciences:
Glad to be of help. I'll be working with your code so I'll let you know if I find something else.

GoldSciences:
I've added another bug fix. This one was spotted after I added initialization of Texture::device to nullptr to find out in shadowmappingomni that it was never being assigned yet it was being used to call vkDestroyImage or something from Texture::destroy().

SaschaWillems:
Closing due to inactivity.

PR #305, opened 2017-03-18
fix error build

ilgrad:
add "#include <functional>" , because  -  error: ‘function’ is not a member of ‘std’

PR #306, opened 2017-03-18
Add missing <functional> include.

Ristovski:
`<functional>` is needed for `std::function` as per http://en.cppreference.com/w/cpp/utility/functional/function.
This fixes a compile error with GCC (_gcc (Gentoo 7.0.0_alpha20170219) 7.0.0-alpha20170219 20170219 (experimental)_)

### DIFF
--- a/base/threadpool.hpp
+++ b/base/threadpool.hpp
@@ -11,6 +11,7 @@
 #include <queue>
 #include <mutex>
 #include <condition_variable>
+#include <functional>
 
 // make_unique is not available in C++11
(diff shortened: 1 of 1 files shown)


SaschaWillems:
I guess this is a duplicate of #305.

Ristovski:
Whoops, missed it since the other PR had a wonky non-descriptive title. (Although "Fix error build" sounds like something I should have looked at.)

SaschaWillems:
Fixed with merge of #306.

PR #308, opened 2017-03-24
Single-threaded rendering mode for the multithreading example

kvark:
The logic of updating and rendering is separated. An ability to update in parallel but then converge to a single thread for rendering is added. This is useful to demonstrate the ability of Vulkan to record commands in parallel and benchmark the effectiveness of such recording (spoiler: it's not obvious).

### DIFF
--- a/multithreading/multithreading.cpp
+++ b/multithreading/multithreading.cpp
@@ -75,13 +75,17 @@ public:
 	// Max. number of concurrent threads
 	uint32_t numThreads;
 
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Thx for the PR. Adding simple single thread toggle is a great addition. Will merge once I got time to review the changes.

SaschaWillems:
Closing this for now. I'll have to rework the whole synchronization part of my samples, which will force me to redo a large part of this sample.

PR #315, opened 2017-04-10
VirtualKeyCode enum and key input handling improvements.

Unarmed1000:
Introduced a platform independent virtual key code enum in the vks namespace.

This could be seen as the first step in making the code less platform dependent.
Using a enum also has the nice feature that modern UI's can do proper code completion.
It also makes sure that all platforms react the same to the keys (meaning less code overall).

This hides the 'keycodes.hpp' so its only included in 'vulkanexamplebase.cpp' it still functions as a 'basic platform code' translation utility.
The 'keyPressed' method was replaced by "void onKeyEvent(const vks::VirtualKeyCode virtualKey, const bool isPressed, const uint32_t rawKeyCode)' which allows for more advanced key handling.

A obvious next step is to do the same for mouse/touch and gamepad functionality which then in turn will make it easier to improve the platform dependent code.

### DIFF
--- a/base/VirtualKeyCode.hpp
+++ b/base/VirtualKeyCode.hpp
@@ -1,5 +1,5 @@
 /*
-* Platform indepdenent virtual key codes
+* Platform independent virtual key codes
 *
 * Copyright (C) 2017 by Sascha Willems - www.saschawillems.de
 *
--- a/base/VirtualKeyCode.hpp
+++ b/base/VirtualKeyCode.hpp
@@ -1,0 +1,86 @@
+/*
+* Platform indepdenent virtual key codes
+*
+* Copyright (C) 2017 by Sascha Willems - www.saschawillems.de
+*
+* This code is licensed under the MIT license (MIT) (http://opensource.org/licenses/MIT)
+*/
+
+#pragma once
+#ifndef VKS_BASE_VIRTUALKEYCODE
+#define VKS_BASE_VIRTUALKEYCODE
+
+namespace vks
+{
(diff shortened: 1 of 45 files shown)


SaschaWillems:
Thx for this PR and all the effort you have put into this. 

My spare time which is currently very limited so I can't tell when I'll be able to take a look at this, as this is a pretty huge PR that I don't want to blindly merge.

Unarmed1000:
Since I have some time at the moment I will go ahead and prepare similar changes for GamePad and Mouse events. This effort should make it much easier to do platform changes in the future at least.

Unarmed1000:
Also if you plan on using GLFW which some of the other pull requests and comments indicate it would be very beneficial to do a change like this first.

The change might also seem huge but its mostly because I replace the keypressed method in all samples. The meat of it really is in the VulkanExampleBase. I basically just went through all the platform dependent key handling code and made sure that anything that reacts to keys do it on the virtual keys so all platforms behave the same. So the base code that reacts to keys where moved into handleKeyEvent instead of being done in each platform section.

SaschaWillems:
> Also if you plan on using GLFW which some of the other pull requests and comments indicate it would be very beneficial to do a change like this first.

Just fyi, I don't have plans to use GLFW, but I'm currently looking into changing the UI for e.g. toggling example options on touch based devices.

SaschaWillems:
As I'm preparing for a cleanup of the repo I'm closing this for now. I may at some point return to this for a bigger UI and cleanup.

PR #318, opened 2017-04-10
Platform independent mouse handling.

Unarmed1000:
Introduced a VirtualMouseButtonFlags struct in the vks namespace. It represents virtual mouse button states.
Currently Android is still handled different as it has vastly different actions being tanken on touch.

Removed all the duplicated mouse handling code and moved it into
    void handleMouseButton(const vks::VirtualMouseButtonFlags& buttonFlags, const glm::vec2& position);
    void handleMouseMove(const vks::VirtualMouseButtonFlags& buttonFlags, const glm::vec2& position);
    void handleMouseWheel(const float delta);

Added custom mouse event handling methods for examples that want to handle mouse events.
    virtual void onMouseButton(const vks::VirtualMouseButtonFlags& buttonFlags, const glm::vec2& position);
    virtual void onMouseMove(const vks::VirtualMouseButtonFlags& buttonFlags, const glm::vec2& position);
    virtual void onMouseWheel(const float delta);

### DIFF
--- a/base/VirtualMouseButtonFlags.hpp
+++ b/base/VirtualMouseButtonFlags.hpp
@@ -1,0 +1,54 @@
+/*
+* Platform independent virtual mouse button flags
+*
+* Copyright (C) 2017 by Sascha Willems - www.saschawillems.de
+*
+* This code is licensed under the MIT license (MIT) (http://opensource.org/licenses/MIT)
+*/
+
+#pragma once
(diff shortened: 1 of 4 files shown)


Unarmed1000:
I am not 100% sure if any of the supported platforms gets their 'mouse' events from a different thread, if so I would recommend that we utilize a proper guarded message queue as a intermediary.

PR #323, opened 2017-04-17
Updated tessellation text interface

sjfricke:
Examining the tessellation code I realized there was a built-in toggle for the wireframe which is hidden due to lack of text overlayed so quickly added it.


SaschaWillems:
Thx 👍

PR #337, opened 2017-05-26
Fixed build failed for _DIRECT2DISPLAY

xiphos:
for _DIRECT2DISPLAY it should not depended on xcb.
fixed by adding macro switch


SaschaWillems:
Thx 👍

PR #340, opened 2017-06-01
WSI Direct2Display

ChaithanyaSingamala:
*Added macro to use default display plane, instead of query for best display plane.
"vkGetPhysicalDeviceDisplayPlanePropertiesKHR" api crashes on some target, even if vkGetPhysicalDeviceDisplayPlanePropertiesKHR is valid.
macro USE_DEFAULT_DISPLAY_PLANE_0 avoid this function call and uses default plane 0(as if display attach at least one plane need to be present with id 0)
by default macro is not enable, if your target crashes for "vkGetPhysicalDeviceDisplayPlanePropertiesKHR" function call, can try enabling USE_DEFAULT_DISPLAY_PLANE_0 macro in VulkanSwapChain.hpp.

### DIFF
--- a/base/VulkanSwapChain.hpp
+++ b/base/VulkanSwapChain.hpp
@@ -108,7 +108,7 @@ public:
 		ANativeWindow* window
 #else
 #ifdef _DIRECT2DISPLAY
-	uint32_t width, uint32_t height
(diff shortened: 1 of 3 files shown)


SaschaWillems:
Thx for your PR. Please give me some time to evaluate the changes before I merge this.

PR #342, opened 2017-06-05
Add support for iOS and macOS via MoltenVK

billhollings:
This PR adds support for running these examples on iOS and macOS using MoltenVK.

Some examples currently do not run successfully on either or both platforms...but this PR gets the core functionality in place...and provides a base for future fixes to both MoltenVK and to individual examples to allow them to run on iOS and macOS.


SaschaWillems:
Thank you very much for this awesome PR. I don't have any Apple devices to test this, so this is very much appreciated and  great addition 👍 

Please give me some time to review it though, before I'll merge this into master 😉

eloj:
Please consider any licensing issues. Some of the headers include verbiage like "If you read this you agree to a license elsewhere".

billhollings:
@eloj 

Good point. Sorry about that...it was unintentional...and a result of starting with files from the **MoltenVK** demos to build the **SW** examples.

I'll modify the copyright notices to an MIT license to bring them in line with the other **SW** files.

@SaschaWillems 

I have an update to this PR on its way anyway. It will include a number of fixes to the examples...so that many more examples will be working out of the box. I will include the copyright changes in that update too.

So...perhaps you can hold off merging this PR until I've updated it. It should be ready sometime early next week.

Regarding the logo...we would like to include it...to clarify how the examples work on _iOS_ and _macOS_...and to recognize Brenwill's contribution. But...if you need to have a policy about that...we understand...and will remove it. The important thing is to make the examples available on _iOS_ and _macOS_. Let me know.

SaschaWillems:
I agree with @eloj on the copyright issue. Licenses throughout the repository should match and changing them to MIT would be appreciated.

And I'm fine with having the logo in the readme for the MacOS and iOS samples so people can see that this is a Brenwill contribution. I'll also add a note about your contributions in the main readme once merged. Credit where credit is due 😉 

One thing I noticed is the size of the assimp libraries. The repo is already pretty big, so is there any way of getting these a bit smaller?

billhollings:
I've addressed the concerns above regarding copyright, logos, and the large AssImp libs. I've also updated the docs, and fixed a number of the examples that were not working on iOS and macOS.

This should be good to go now. Let me know if anything else needs to be addressed.

SaschaWillems:
Thanks for the changes! Looks great so far, will take a closer look at this later this weekend and merge it into master 👍

SaschaWillems:
Just out of curiosity: For all other platforms you need to [enable the platform specific instance extension](https://github.com/SaschaWillems/Vulkan/blob/master/base/vulkanexamplebase.cpp#L31), but I don't see  the ones for iOS or MacOS in there. Aren't those required, or is MoltenVK adding them implicitly?

billhollings:
MoltenVK doesn't actually validate extensions right now...but it should...and I'll add it to the MoltenVK todo list.

In the meantime...I'll add the extension enabling to the SW examples...as you indicate...and I'll issue another PR.

billhollings:
I've submitted PR #349 to add the MoltenVK extension names to vkCreateInstance().

PR #343, opened 2017-06-14
Basic support for OS X using MoltenVK

renderdude:
Adds basic support on OS X using MoltenVk. UI is handled by glfw, but I haven't add any callback for key presses and mouse movement (yet).
All the examples compile, but only about one third actually work for several reasons:
1. MoltenVk doesn't currently support VkCmdDispatch
2. MoltenVk doesn't currently support compute/geometry/tesselation shaders
3. There's a bug in the shader conversion (glsl/spir-v -> Metal) in that it transforms MatrixInverse in SPIR-V to inverse in Metal. ASAIK, there is no inverse in the Metal shading langugae.


renderdude:
Oops. Didn't see that someone else had the same idea. Oh well, discard if it's a duplication.

SaschaWillems:
PR #342 is actually by one of the MoltenVK developers and also adds iOS along with MacOSX. So yes, this seems like a duplicate.

SaschaWillems:
I just merged #342 by one of the MoltenVK developer. That PR adds support for macOS and iOS.

PR #347, opened 2017-06-27
Proper app termination after WM_QUIT

aleksysokolowski-intel:
According to MSDN PeekMessage returns false only when the message queue is empty. If there is another message after WM_QUIT, the part of the code responsible for proper closing of the application would never be triggered. This change handles WM_QUIT message properly.


SaschaWillems:
Thx for the PR 👍

PR #352, opened 2017-07-04
Support installing demos, support out-of-tree builds

jku:
Hi,
I'm using your demos (at least the parts without ambiguous licensing) as a Vulkan smoke testing tool in Yocto. The use-case requires installing so I cooked up this patch.

A define for install path is a bit unusual: I did because I didn't want to break the developer use case of "compile &run without installing" by default... There's no install targets at all if the define is not set: this way no-one can install binaries that would try to load data from CMAKE_SOURCE_DIR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,6 +16,8 @@ include_directories(base)
 OPTION(USE_D2D_WSI "Build the project using Direct to Display swapchain" OFF)
 OPTION(USE_WAYLAND_WSI "Build the project using Wayland swapchain" OFF)
 
+set(RESOURCE_INSTALL_DIR "" CACHE PATH "Path to install resources to (leave empty for running uninstalled)")
+
 # Use FindVulkan module added with CMAKE 3.7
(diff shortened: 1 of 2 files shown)


jku:
Based on the initial windows build failure I've modified getAssetPath() to do what it used to do if the path isn't set by the build system: I don't really have knowledge to do anything else with windows

SaschaWillems:
> (at least the parts without ambiguous licensing)
Can you elaborate on this? I'd like the licensing to be as clear as possible, so any if there's any licensing ambiguity I'll try to fix that.

And thx for the PR. Please give me some time to review it.

jku:
Oh the code licensing seems completely straight forward: I'm referring to the models that you mention in the README:  

> Please note that (some) models and textures use separate licenses. Please comply to these when redistributing

Yocto/openembedded itself doesn't distribute them but we still try to keep to open source only: I'm opting for safe rather than sorry so I'm removing things on your list if I couldn't find a license that is "open source" enough.

PR #365, opened 2017-08-09
Fixed pure virtual function buildCommandBuffers in VulcanExampleBase

RyperHUN:
The comment says, it is a pure virtual function, but it was not. This was it was easy to make a mistake


SaschaWillems:
I'm preparing for a repo cleanup, so I'm closing this. But I plan on cleaning up the base code and may come back to this at a later point.

PR #367, opened 2017-08-17
Build base/*.cpp as a library to prevent recompiling it multiple times

jpbruyere:
modifié :         CMakeLists.txt
	nouveau fichier : base/CMakeLists.txt

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,12 +82,14 @@ endif(CMAKE_COMPILER_IS_GNUCXX)
 add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 add_definitions(-std=c++11)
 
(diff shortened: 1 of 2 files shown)


jpbruyere:
failed with gcc on travis => closed

SaschaWillems:
It looks like there may be just a define missing (xcb). Moving base to a shared library is something I wanted to work on multiple times but never actually got around doing so. I'll try to take a look at why this only fails with gcc on travis (and works fine with clang).

jpbruyere:
I've enabled travis for my fork, I'll make some tests. But for win and android builds I can't check that it's working fine.

jpbruyere:
Now I think it's ok, It compiles fine. I'm trying to reopen the pull request but I've reset several commits and make some forced pushes. Maybe I'll have to make another pull request. let's try...

PR #369, opened 2017-08-20
add datas and shaders as custom targets

jpbruyere:
This PR would add custom targets for compiling shaders and copying datas so that each time a shader or a data file is changed, it will be included in the build.

- Add custom target for data copy to the build dir but only if build is done outside source dir.
- Add custom targets for shader compilation, one per example
- glslc compiler is searched and build is canceled if not found

precompiled spir-v shaders could be removed from git.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -82,7 +82,64 @@ endif(CMAKE_COMPILER_IS_GNUCXX)
 add_definitions(-D_CRT_SECURE_NO_WARNINGS)
 add_definitions(-std=c++11)
 
-file(GLOB SOURCE *.cpp )
+# Find glslc shader compiler.
+# On Android, the NDK includes the binary, so no external dependency.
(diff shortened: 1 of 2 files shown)


SaschaWillems:
Sry for the late reply. Just took a look at this and noticed a few problems:

I'd prefer to have the custom targets to be optional. Right now it looks the change will fail the whole CMAKE script if glslc is not present:

    IF (GLSLC-NOTFOUND)
        message(FATAL_ERROR "Could not find glslc compiler!")
    ELSE()
        message(STATUS "Found spir-v compiler: " ${GLSLC})
    ENDIF()

So for those users that don't have the SDK installed (which currently is not a requirement ) this won't work at all.

This PR also includes changes to the indentation of the vulkanexamplebase source, while the PR itself is actually only about the cmake file.

SaschaWillems:
Closing due to inactivity.

PR #375, opened 2017-09-12
pSparseMemoryRequirementCount variable initialization

aleksysokolowski-intel:
According to spec: if pSparseMemoryRequirements is not NULL -> pSparseMemoryRequirementCount must point to a variable set by the user to the number of elements in the pSparseMemoryRequirements array.


SaschaWillems:
Thx 👍

PR #378, opened 2017-09-21
Adds specialization constant to computeheadless

chuggafan:
I noticed that you had the comment in headless.comp for the computeheadless project, and as such I added specialization constants and removed the comment about said specialization constants


SaschaWillems:
Thx for the PR!

PR #382, opened 2017-10-09
base: Check for swapchain blit source feature instead of destination

laanwj:
As I understand it the goal is to use the swapchain image as a transfer source (for taking screenshots), not as a transfer destination.


SaschaWillems:
Yes, that if-condition is obviously wrong, thx for pointing this out. Changing it to blit source is only a partial solution though as the check should also include the transfer source flag bit for implementations that don't support blits. Will add that too.

PR #388, opened 2017-10-20
Call Android subprocesses with python3

cirosantilli:
Ensure that python 3 is used on systems in which python2 is the default
python executable, but which have both pythons installed.

Tested in Ubuntu 16.04.


cirosantilli:
Wait, I have a better idea, I'll try to make build.py both callable from build-all.py and runnable, this way the shseband python3 will be used automatically for it.

PR #391, opened 2017-10-26
Allow Linux users to build Android even if Python 2 is their default Python

### LINKED ISSUE #297: How do you want to implement the Android build system for Linux? bash, python, cmake, etc.
I'm considering porting the android build to work from Linux.

If I do, will you consider merging it?

Looks trivial, just porting those `.bat` under `android`, which basically just call `mkdir`, `cp` and `android`?

If yes, then how do you want it to be done?

I see a few possibilities:

- bash: would be easy to do, but would duplicate the Windows .bat thing
- python: also easy, and would work both on Windows and Linux. Only possible concern is availability in Windows
- cmake: not sure if possible, a bit less easy, but would be good since already used in repo

cirosantilli:
After this commit, users who have the `python` executable as Python 2 can do simply:

    cd android/
    ./build-all.py
    ./build.py

Previously, the following would fail:

    python3 build-all.py

since build-all.py did subcalls to `python build.py`.

Remove install-all.py as it is redundant with `build-all.py -deploy`,
instead of fixing it as well.

Introduce argparse since we are separating argument parsing out of
the `main()` function.

Tested in Ubuntu 16.04, behaviour should be unchanged for Windows.

Fix #297.


SaschaWillems:
Thx for your effort, this is really appreciated. I'll take a look at these changes this weekend and merge them into the repository if everything works over here.

PR #394, opened 2017-11-09
Ensuring that the host visible buffer to read compute data from is on the VRAM (device)

chakmeshma:
The selected memory type used to create the buffer memory "hostMemory" declared in line 235 was not necessarily corresponding to a device-local memory queue, i added the requirement for a device-local buffer memory.


SaschaWillems:
I don't think this is correct. Your change always checks if the requested memory type supports ```VK_MEMORY_PROPERTY_DEVICE_LOCAL_BIT```, even for the first host visible buffer where this won't work on most discrete GPUs that don't have shared memory.

That's not correct imo, so can you elaborate as to why this change would be correct?

SaschaWillems:
Closing due to inactivity.

PR #400, opened 2017-11-22
VulkanUIOverlay: Use the correct initial layout

jekstrand:
Using an initial layout of VK_IMAGE_LAYOUT_UNDEFINED together with
VK_ATTACHMENT_LOAD_OP_LOAD means that the render pass starts off with
completely undefined contents.  Since we want to draw the overlay on top
of the previously rendered image, we want to use a real layout.  Since
the calling code has already transitioned to VK_IMAGE_LAYOUT_PRESENT_SRC_KHR,
that's the layout we need to use.

This was never caught by validation because VK_IMAGE_LAYOUT_UNDEFINED is
the one layout you're allowed to use as an initial layout at any time
regardless of the image's current layout.

### DIFF
--- a/base/VulkanUIOverlay.cpp
+++ b/base/VulkanUIOverlay.cpp
@@ -337,7 +337,7 @@ namespace vks
 		attachments[0].storeOp = VK_ATTACHMENT_STORE_OP_STORE;
 		attachments[0].stencilLoadOp = VK_ATTACHMENT_LOAD_OP_DONT_CARE;
 		attachments[0].stencilStoreOp = VK_ATTACHMENT_STORE_OP_DONT_CARE;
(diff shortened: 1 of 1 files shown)


jekstrand:
FYI, you really don't want to be transitioning to `VK_IMAGE_LAYOUT_PRESENT_SRC_KHR` as an intermediate on our driver.  Window system images are so restrictive that you take a pretty heavy resolve hit every time you do that.  It would be much better to just leave it in `VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL` between render passes.

SaschaWillems:
Thx for the PR and esp. the insight on the Intel driver's workings. Much appreciated!

PR #401, opened 2017-12-01
Request coherent memory for model staging buffer

JarredHDavies:
There is no guarantee in the spec on the order of the following memory types:

```
VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_CACHED_BIT,
VK_MEMORY_PROPERTY_HOST_VISIBLE_BIT | VK_MEMORY_PROPERTY_HOST_COHERENT_BIT,
```

which means we can't assume that the first memory type that is host visible is going to be coherent.


SaschaWillems:
Thx for your PR!

PR #410, opened 2018-01-10
Fix multisampling example for AMD hardware.

AaronHaganAMD:
The multisampling example is broken on AMD hardware after the UI menu changes. Looks like the resolve was happening twice once for the model and once for the GUI window. This change fixes it on all families of AMD hardware.


SaschaWillems:
Thx for this fix. Much appreciated!

PR #419, opened 2018-02-01
multisampling, subpasses, textoverlay: Fix renderpass initial layout.

### LINKED ISSUE #418: multisampling, subpasses, textoverlay: Incorrect renderpass initial layout
If 2 renderpasses render to same render target, second renderpass must not have this attachment initial layout set to ```UNDEFINED```, but to actual image layout. Otherwise first renderpass result may be discarded.

Spec language:
```
Setting the old layout to VK_IMAGE_LAYOUT_UNDEFINED implies that the contents of the image subresource need not be preserved.
```

iostrowsINTEL:
Fixes #418


SaschaWillems:
Thx for your PR. LGTM. Will test and merge as soon as time allows.

PR #427, opened 2018-02-28
macOS update

yanchith:
This updates include paths in `examples.h`, README for molten, and xcode project files so that the examples can be run on macOS.

Particulary in the xcode files, I am not sure if I did the right thing, so I would appreciate someone checking that.

Some of the examples do not run, but that is documented in `examples.h`, and is because of vulkan/metal mismatch.


SaschaWillems:
Thx for the PR. LGTM, but I don't own any Apple device I could check this on. I'll merge this in the coming days.

Issue #425, opened 2018-02-22
parallax: non-uniform texture fetches


msiglreith:
The parallax shader uses samples the normal map in a non-uniform branch (https://github.com/SaschaWillems/Vulkan/blob/master/data/shaders/parallax/parallax.frag#L73), which procudes invalid SPIR-V up to my understanding: SPIR-V Spec, 2.20 Code Motion:

> Texturing instructions in the Fragment Execution Model that rely on an implicit derivative cannot be moved into control flow that is not known to be uniform control flow within each derivative group.

cforfang:
Assuming the texture doesn't have mipmaps, which I believe is the case here, can the texturing instruction be said to "rely on an implicit derivative"? If not, it _seems_ OK, but I don't actually know how SPIR-V works here...

I suppose a fix, if needed, would be to use ``textureLOD()`` with lod=0 instead?

cforfang:
Also, taking a second look, is that actually a non-uniform branch? All the control flow seems to be based on values in UBOs, so should be draw-call uniform right?

msiglreith:
@cforfang The break condition depends on the read texture at non-uniform texture coordinates. Therefore it appears to me that the control flow is non-uniform for the executed primitive group.

cforfang:
Ah, right, yes -- I completely missed that :) I still don't quite get the wording in the SPIR-V spec when it says "cannot be moved". Who's moving anything? I assumed the spec is meant to describe what must be true in valid SPIR-V, but this almost sounds like a directive to compilers.

By contrast, if it said something like "Texturing instructions in the Fragment Execution Model that rely on an implicit derivative **cannot be placed in** control flow that is not known to be uniform control flow within each derivative group." that seems pretty clear, but as-written it almost sounds more like it's meant to tell compilers it should not move texturing-instructions around without considering uniform control flow during compilation.

In any case, to be 100% sure this is legal I guess {{textureLod()}} is the way to go.

msiglreith:
Agree, opened an issue on the official repository: https://github.com/KhronosGroup/SPIRV-Headers/issues/58

cforfang:
Ah, sorry -- didn't see your edit that you were going to do it :) I agree, please check if #431 looks OK.

msiglreith:
Fixed in https://github.com/SaschaWillems/Vulkan/pull/431

PR #431, opened 2018-03-07
Parallax: fix texture() in non-uniform control-flow

cforfang:
Fix for issue #425. I changed all use of `sNormalHeightMap` instead of trying to parse out where exactly the control-flow is non-uniform (shouldn't make any difference with 1 mip level anyway).

Warning: I haven't actually tested the code yet :)


SaschaWillems:
Thx for this PR!

PR #432, opened 2018-03-07
Fix capitalization of glslangValidator

cforfang:
compileshaders.py didn't work for me on Linux when ``glslangvalidator`` is all lowercase.


SaschaWillems:
Thx for this PR!

PR #437, opened 2018-03-19
Use transient depth/stencil buffer

gary-sweet:
We don't need to store the d/s buffer when its load-op
is configured to clear. This can also save bandwidth and
increase performance slightly.

Flagging the depth/stencil buffer as transient and using lazy
allocation allows some implementations (tile-based renderers)
to avoid allocating the buffer at all in most cases since it
now never needs to be stored.


SaschaWillems:
Sorry for the late reply, but I just tested this and it won't work on GPUs that don't offer lazily allocated memory. 

In order to get this working on all platforms there needs to be check upfront to see if that memory type is supported. I'll try to take a look at that and then also implement your changes.

gary-sweet:
> won't work on GPUs that don't offer lazily allocated memory
Ah yes, sorry about that. 

> I'll try to take a look at that and then also implement your changes
That's great, thanks!

PR #455, opened 2018-04-08
Small fix in Push descriptors example

hiddenbitious:
Check if pointer to vkGetPhysicalDeviceProperties2KHR is not null.


SaschaWillems:
The second function pointer check is definitively wrong. Will fix that. Thx for noticing!

PR #456, opened 2018-04-13
Raytracing - fixed problem with synchronization of uniform data updates.

Ekzuzy:
Uniform buffer data in the Raytracing sample was updated while compute shader, which was using it, was being executed. This Pull Request fixes synchronization problems: changes time at which uniform buffer data is updated and adds buffer memory barrier into a command buffer.

### DIFF
--- a/examples/raytracing/raytracing.cpp
+++ b/examples/raytracing/raytracing.cpp
@@ -283,6 +283,7 @@ public:
 
 		VK_CHECK_RESULT(vkBeginCommandBuffer(compute.commandBuffer, &cmdBufInfo));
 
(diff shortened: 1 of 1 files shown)


cforfang:
Since the UBO is `VK_MEMORY_PROPERTY_HOST_COHERENT_BIT` I don't think you need a pipeline-barrier here. The memory is made visible and available on the GPU when you call `vkQueueSubmit` (see "6.9 Host Write Ordering Guarantees" in the spec).

Ekzuzy:
UBO has a `VK_MEMORY_PROPERTY_HOST_COHERENT_BIT` memory bound so there is no need to call the vkFlushMappedMemoryRanges(). But I forgot that vkQueueSubmit() makes memory writes writes implicitly available.

PR #460, opened 2018-04-15
Replace hardcode value with VkColorComponentFlagBits enum

sjfricke:
to keep the consistency, I added the [VkColorComponentFlagBits](https://www.khronos.org/registry/vulkan/specs/1.0/html/vkspec.html#VkColorComponentFlagBits) enums instead of just hardcoding the `0xf` for both good coding pratice saving people a google search of what `0xf` represents

### DIFF
--- a/examples/bloom/bloom.cpp
+++ b/examples/bloom/bloom.cpp
@@ -680,7 +680,7 @@ public:
 
 		VkPipelineColorBlendAttachmentState blendAttachmentState =
 			vks::initializers::pipelineColorBlendAttachmentState(
-				0xf,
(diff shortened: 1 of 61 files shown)


SaschaWillems:
Not sure about this (was suggested multiple times). Didn't do this from the start because I don't like that long flag combination line myself and would rather define this as an initializer and reuse that.

sjfricke:
No thats fair, Its part my fault for not looking if this was mentioned before making PR. I guess it just seemed odd since everything is so verbose as it is to decide this one thing shouldn't use the long enums syntax. I guess maybe a better approach would be to make an `#define` out of it.

PR #472, opened 2018-04-27
Update VulkanTexture.hpp, fixed assignment

elect86:


elect86:
Hi, 

`imageLayout = imageLayout;` doesn't make any sense, so I thought to assign it to the class variable..

SaschaWillems:
I see, that's wrong, though it's in the linear tiling path that's never actually used. Will push a correct fix for this.

PR #478, opened 2018-05-06
Fixed android shadowmapping crash

sjfricke:
Android shadowmapping crashed when switching between models

Was getting `vulkanExample: Could not load mesh from "models/samplescene.dae"!` and realized that `samplescene.dae` was never being loaded into the android build


sjfricke:
Issue still relevant, just redoing to adapt to new Android layout

PR #480, opened 2018-05-08
Fix possible freeze on "VK_ERROR_OUT_OF_DATE_KHR"

zaps166:
NVIDIA 396.18.08 fixes Alt+Tab issue when "Allow Flipping" option is enabled (X11 Compositor: OFF).
All Vulkan examples can still freeze X11 on Alt-Tab even after driver bugfix. After swap chain recreation we must call `vkAcquireNextImageKHR` again - fixes the X11 and application freeze after pressing Alt-Tab.

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -608,14 +608,17 @@ void VulkanExampleBase::updateOverlay()
 
 void VulkanExampleBase::prepareFrame()
 {
-	// Acquire the next image from the swap chain
-	VkResult err = swapChain.acquireNextImage(semaphores.presentComplete, &currentBuffer);
-	// Recreate the swapchain if it's no longer compatible with the surface (OUT_OF_DATE) or no longer optimal for presentation (SUBOPTIMAL)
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Can you elaborate a bit? This will affect all platforms and I'm not sure if that for..loop will do any good outside of that Alt+Tab issue.

zaps166:
The loop is needed to acquire the image again after recreating swap chain:
```
1. `acquireNextImage`
2. if return value is `VK_ERROR_OUT_OF_DATE_KHR` or `VK_SUBOPTIMAL_KHR` {
    3. recreate swapchain
    4. `acquireNextImage` (do it again, this is the bugfix)
    5. if return value is error (we still have an error) {
        6. display error
    }
    7. return
}
8. else (if return value is OK) {
    9. return (break the loop here)
}
```

Or just write:
`VK_CHECK_RESULT(swapChain.acquireNextImage(semaphores.presentComplete), &currentBuffer);` 
after:
`windowResize();`

SaschaWillems:
Can you check if this fix is still necessary with current NVIDIA drivers? I'm preparing for a cleanup of the repo and want to close all open PRs beforehand.

zaps166:
Nvidia 396.54 driver doesn't freeze without this patch.

PR #489, opened 2018-05-21
Disable fragment shader when generating shadow map.

prideout:
This fixes the following validation warning:

fragment shader writes to output location 0 with no matching attachment


SaschaWillems:
Thx for the PR. Very much appreciated!

PR #497, opened 2018-06-29
Don't set an object name for the wireframe pipeline when fillModeNonSolid is not supported

JarredHDavies:
Fixes segfault in debugmarker example on platforms where fillModeNonSolid is not supported.


SaschaWillems:

PR #499, opened 2018-07-10
Fix for multisample depth resolve attachment and storeOp for the MSAA color attachment

AndreasReiten:
Removed the attachment intended for resolved depth, which Vulkan can't do (yet?). The storeOp for the MSAA color attachment can safely be changed to VK_ATTACHMENT_STORE_OP_DONT_CARE as it is resolved before the end of the subpass.

### DIFF
--- a/examples/multisampling/multisampling.cpp
+++ b/examples/multisampling/multisampling.cpp
@@ -248,13 +248,13 @@ public:
 	{
 		// Overrides the virtual function of the base class
 
-		std::array<VkAttachmentDescription, 4> attachments = {};
+		std::array<VkAttachmentDescription, 3> attachments = {};
 
 		// Multisampled attachment that we render to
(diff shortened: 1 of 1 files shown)


chrisvarns:
Don't the validation layers complain if you don't provide a depth resolve target, even when it's not used? Maybe they fixed it.

SaschaWillems:
Thx. Looks good so far. Current layers no longer complain. I'll do some tests on different platforms and then merge if it works everywhere.

examples/multisampling/multisampling.cpp
```diff
@@ -248,13 +248,13 @@ class VulkanExample : public VulkanExampleBase
 	{
 		// Overrides the virtual function of the base class
 
-		std::array<VkAttachmentDescription, 4> attachments = {};
```
skurwielstein: 3

PR #502, opened 2018-07-12
Add missing flush in one of the buffer creation helpers.

sg-brcm:
This caused gears to break, after a driver change removed an implicit flush in our buffer copy implementation that had been hiding this issue. I could have just changed gears to use coherent memory, but it seemed sensible to fix the helper infrastructure anyway.


SaschaWillems:
Thx for the PR! Very much appreciated.

PR #516, opened 2018-11-03
Update linux build instructions with required dependencies

apnavik:


apnavik:
Oh, Yes, I missed it.

PR #517, opened 2018-11-06
Fix incorrect rendering of right eye on radv

pendingchaos:
I'm not too familiar with the spec regarding multiview, so perhaps this could be a radv bug.

Sometimes the error isn't noticeable until you tweak the eye separation.

This error occurs with both a Vega 64 and a RX 580.

### DIFF
--- a/examples/multiview/multiview.cpp
+++ b/examples/multiview/multiview.cpp
@@ -345,7 +345,7 @@ public:
 			framebufferCI.pAttachments = attachments;
 			framebufferCI.width = width;
 			framebufferCI.height = height;
-			framebufferCI.layers = 1;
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Can you please check this with current code? The multiview sample had a bug that made it use a wrong render pass that could have caused this issue too.

pendingchaos:
With the current code in master, the right eye still renders incorrectly.

After applying this PR's change, it renders correctly.

pendingchaos:
Seems the Vulkan spec requires the framebuffer's layer count to be 1 when multiview is used.

A fix is at https://gitlab.freedesktop.org/mesa/mesa/merge_requests/20

PR #530, opened 2018-12-18
wayland: Port to xdg-shell stable

soreau:
Many wayland compositors do not support the deprecated wl_shell protocol but support xdg-shell protocol. This means that the vulkan demos fail to run in such environments. This changes the protocol used from the deprecated wl_shell to the xdg-shell stable protocol, which is the protocol clients are expected to use and compositors are expected to implement.


SaschaWillems:
Can you please add some description as to what this PR does? Will it break something, or will it still work with ever wayland distro?

soreau:
If you look [here](https://people.freedesktop.org/~whot/wayland-doxygen/wayland/Server/page_iface_wl_shell.html) for instance, you can see that wl_shell is a deprecated protocol. I do not have a list of what compositors support xdg-shell stable but I think gnome and sway do and I know wayfire does.

soreau:
Additionally, "Applications are recommended to move on to the stable xdg-shell version." from [a post](https://lists.freedesktop.org/archives/wayland-devel/2018-November/039670.html) by the lead wayland developer.

soreau:
- Rebased on master
- Removed newline whitespace problems

SaschaWillems:
Thanks for the clarification and your PR. Looks good to me 👍

Issue #526, opened 2018-12-06
Compute Cull and LOD - Frustum culling broken


a-day-old-bagel:
There's some problem in the frustum culling math in the compute cull and lod example. This happens to me on both Windows and Linux (xlib), using a 980ti, and I thought I remembered it working before, which is odd since it doesn't look like this code has changed in a while. I can only explain the problem with a video, so here is a GIF of what happens:
https://gfycat.com/PoshRemarkableHammerheadshark

SaschaWillems:
Noticed that too on different vendors. I'll try to find out why this happens and how to fix it.

sirrodgepodge:
the day I understand this issue I will be so proud

SaschaWillems:
Fixed with #533.

wumo:
The frustum created from a matrix is not right for vulkan's clip coordinate system as the depth is in [0,1] not [-1,+1].
The [original](https://www.gamedevs.org/uploads/fast-extraction-viewing-frustum-planes-from-world-view-projection-matrix.pdf) derivation of near frustum plane for OpenGL assuming the z'/w' should be in [-1,+1], but the correct formula should follows Direct3D's near plane calculation which assumes z'/w' in [0,1].
So code:
https://github.com/SaschaWillems/Vulkan/blob/8526c204e78ff517cb227ca1cd5a5d2be7b255e9/base/frustum.hpp#L48-L51
should be changed to:
```
planes[FRONT].x =matrix[0].z; 
planes[FRONT].y = matrix[1].z; 
planes[FRONT].z = matrix[2].z; 
planes[FRONT].w = matrix[3].z; 
```

PR #533, opened 2018-12-27
Pass the full count of the indirectCommands to vkCmdDrawIndexedIndirect

birsoyo:
Items with zero instance count still need to be processed.

This fixes the problem reported with #526.


SaschaWillems:
Thx for this PR! Very much appreciated :)

PR #537, opened 2019-01-20
fix: member not set because parameter same to member name

axilesoft:


axilesoft:
>           Thx for spotting this and supplying a PR. Much appreciated.

You are welcome! And you may also change the comment text of the parameters.

PR #542, opened 2019-02-04
Fix Android build error.

prideout:
The latest version of Android Studio fails with:

    Could not find com.android.tools.build:aapt2

To use AAPT2, we must now have a dependency on the `google()` repo in both places in the root build.gradle file, which is documented [here](https://developer.android.com/studio/releases/#aapt2_gmaven).


SaschaWillems:
Thx for your PR! Very much appreciated :)

PR #549, opened 2019-02-23
Make FPS counter more accurate

djdeath:
While trying to figure a discrepancy between the FPS counter from the
overlay we've introduced in Mesa [1] and the counter in the Vulkan
demos, I figured the demos are not accounting for part of the
rendering loop but rather just the amount of time spent rendering.

This changes accounts for the total amount of time between 2 frames. I
don't think any difference is visible until you reach high frame rates
of 100s or so.

[1]: https://gitlab.freedesktop.org/mesa/mesa/merge_requests/303

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -250,7 +250,7 @@ void VulkanExampleBase::renderFrame()
 			timer -= 1.0f;
 		}
 	}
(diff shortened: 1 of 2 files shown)


SaschaWillems:
Thanks for your PR, very much appreciated.

Measuring the whole render loop actually does make more sense :)

PR #554, opened 2019-03-06
Fix sample count limits in deferredmultisampling

gary-sweet:
The sample count was previously fixed at 8, which is not supported on some platforms. The code now checks for the maximum supported sample count and uses that instead.
    
If you would prefer the upper limit to be fixed at 8, that is obviously trivial to change.


gary-sweet:
This has an extra merge that wasn't meant to be here. I'll try again.

PR #555, opened 2019-03-06
Fix sample count limits in deferredmultisampling

gary-sweet:
The sample count was previously fixed at 8, which is not supported on some platforms. The code now checks for the maximum supported sample count and uses that instead.

If you would prefer the upper limit to be fixed at 8, that is obviously trivial to change.


gary-sweet:
I'm not sure why the appveyor test is failing. I don't think it's related to my change.

SaschaWillems:
Thx for your PR, will take a look at it and merge later this week.

And don't worry about the appveyor failure, this has stopped working for me some time ago and haven't had the time to check why.

SaschaWillems:

PR #596, opened 2019-07-26
Fix a couple of small errors in ray tracing examples

jmacnak-nv:
nv_ray_tracing_basic.cpp
 - update to 3 component vertex data (4 component isn't supported)
 - update type to TOP_LEVEL when building the TLAS

nv_ray_tracing_reflections.cpp
 - update AS info type to TOP_LEVEL when building the TLAS

nv_ray_tracing_shadows.cpp
 - update AS info type to TOP_LEVEL when building the TLAS

Validation layer changes which will surface these errors coming soon :)


SaschaWillems:
Thanks for your PR. Very much appreciated, and also thx for your raytracing contributions to the SDK :)

Issue #600, opened 2019-08-20
Validation error in pipeline statistics example


cos-public:
There is a validation error in pipelinestatistics example:
ERROR: [Validation] Code 0 :  [ VUID-vkCmdBeginQuery-queryType-00800 ] Object: 0x1f1bab34e70 (Type = 6) | VK_QUERY_CONTROL_PRECISE_BIT provided to vkCmdBeginQuery, but precise occlusion queries not enabled on the device. The Vulkan spec states: If the precise occlusion queries feature is not enabled, or the queryType used to create queryPool was not VK_QUERY_TYPE_OCCLUSION, flags must not contain VK_QUERY_CONTROL_PRECISE_BIT (https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkCmdBeginQuery-queryType-00800)

BobLChen:
set to zero
https://github.com/SaschaWillems/Vulkan/pull/601

SaschaWillems:
Setting the flags to zero is not the right solution. The right solution is to check if the device supports precise queries and then enable and use that. I'll take a look at that.

PR #601, opened 2019-08-27
fix incorrrect flags

BobLChen:


BobLChen:
https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-vkCmdBeginQuery-queryType-00800

SaschaWillems:
I guess this is related to: https://github.com/SaschaWillems/Vulkan/issues/600?

If so, then the change from this PR wouldn't be an ideal solution. The correct way would be to query device capabilities and select the flags accordingly.

I'll take a look at that.

SaschaWillems:
Thanks for your and hinting me at this error.

This was also wrong in the tessellation terrain demo, and i have fixed both.

PR #606, opened 2019-09-05
multisampling: only build the per-sample pipeline if the device supports it

chrisforbes:
Fixes: #605


SaschaWillems:

Having this pipeline optional also requires a slight change to the UI, I'll add that too.

PR #609, opened 2019-09-06
Recompile shaders with more modern glslang

chrisforbes:
The SPIRV previously checked in had various issues -- most commonly, it
would use the VertexId and InstanceId builtins rather than the
replacmenet VertexIndex and InstanceIndex builtins required by Vulkan.


SaschaWillems:
Thansk for your PR. Will try to merge this sometime next week.

I actually wanted to replace the compiled shaders with runtime compilation using glslang, but as usual the Android side of things is stopping me from doing this, and I don't really have the time to fix  my Android build setup to work with glslang :/

chrisforbes:
Friendly ping

PR #615, opened 2019-09-17
Added routine for reading shader code from embedded buffer.

dov:
- New function loadShaderFromBuffer() for loading a shader from an
  inmemory string.
- New function readBinaryFile() for reading a binary file.
- Rewrote loadShader() to use the above two functions.


SaschaWillems:
As there is no sample in my repo that requires such a function I'll close this PR for now.

PR #632, opened 2019-11-21
Fix displacement sample blindly enabling BC compression feature

chrisforbes:
The loadAssets() function already supported whichever compression feature was supported, but the features enabling was blindly using BC. This causes trouble on Android devices which typically do not support BC.


chrisforbes:
@SaschaWillems friendly ping

SaschaWillems:
Sorry, haven't had time to check this yet. Been mostly on business trips etc., but will try to take a look at this during the weekend.

PR #635, opened 2019-11-25
Update ssao.frag

liuhongyi0101:
samplePos.z bias  can remove banding

### DIFF
--- a/data/shaders/ssao/ssao.frag
+++ b/data/shaders/ssao/ssao.frag
@@ -40,6 +40,8 @@ void main()
 
 	// Calculate occlusion value
 	float occlusion = 0.0f;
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Thanks for your PR. Could you fix spelling? It should be spelled bias.

liuhongyi0101:
full 32 bits of depth

liuhongyi0101:
> Thanks for your PR, very much appreciated.
> 
> Could you fix spelling before I merge? It should be spelled bias.
> 
> And just to check: What depth format did you test with? Banding is fixed for full 32 bits of depth with a fixed bias, but still partially present with 24 bit formats.

full 32 bits of depth

PR #639, opened 2019-12-18
Documentation & Blueprints for each example.

sadernalwis:
Improve the documentation with blueprints in SVG+PNG format, for each of the current and future examples.


SaschaWillems:
Can you elaborate what this is about?

sadernalwis:
i have been educating myself on Vulkan using your repository which has helped me greatly navigate Vulkan through its complexities. And I started to create documentaion on my own projects for reference/adapt of my team who are not really familiar with Vulkan.

Even for me, it gets quite daunting when going deep in to complex Vulkan sub-passes/render-pass/pipeline-caches/memory management etc.

I think using layered-SVG image files, I can help create node-like struct-structures on a particular Vulkan application/pipeline in a platform-agnostic masnner. for Example:

Layer 1 : the positioning/encapsulation of structs (as for the commited offscreen sample, which is still very incomplete btw)
Layer 2 : Memory Transfers/Threads
Layer 3: Code blocks
Layer 4 : Annotations

even though this is not completely thought out, but i think with more contirbutions and even just error-checking, this can be helpful for even for a complete beginner(without experience on OpenGL) to get started on Vulkan.

Your repository has helped me lot to migrate my work over to Vulkan and i think it's about time i started contributing an a way I can :)

SaschaWillems:
This sounds very interesting, but as soon as this becomes part of the repository, those blueprints need to be maintained, end every time I do relevant changes to one of the samples they would have to be redone. 

So I think it would be a better idea to have them put into a separate repository and link to that from this project's wiki instead of putting them right into this repo.

SaschaWillems:
I'm not going to add this to my sample repository as I'm focusing on the official Khronos repository. Something like this may be added there in the future.

sadernalwis:
> This sounds very interesting, but as soon as this becomes part of the repository, those blueprints need to be maintained, end every time I do relevant changes to one of the samples they would have to be redone.
> 
> So I think it would be a better idea to have them put into a separate repository and link to that from this project's wiki instead of putting them right into this repo.

Hello Sascha,
I sincerely apologize for the very delayed reply. I wasn't able to touch this area of the development until now but gladly I'm back at it again.
I understood what you meant, and yes i will try to maintain the blueprints , and try to continue contribute to this project via my fork.
Khronos repository seems up and coming very excited to see its work.
Best regards,

Issue #645, opened 2020-01-02
computecloth reports lots of validation errors related to synchronization of buffers across queues


jherico:

SaschaWillems:
I'm aware of this, see [this PR on the Khronos Vulkan samples repo](https://github.com/KhronosGroup/Vulkan-Samples/issues/3).

PR #646, opened 2020-01-02
Fix validation and improve sync for computecloth

jherico:
For for #645

The current computecloth example emits lots of validation errors.  Debugging it I found several problems.

* According to the spec, when transferring between queues, there must be both a release barrier in the source queue and an acquire barrier in the destination queue.  The current example doesn't add any barriers to the graphics queue at all.  
* The access masks and pipeline stages don't match up.  `VK_ACCESS_VERTEX_ATTRIBUTE_READ_BIT` can't be used with `VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT`.  
* Inside the `for (uint32_t j = 0; j < iterations; j++) {` loop, the barriers should have both srcQueueFamilyIndex and srcQueueFamilyIndex set to compute, instead of set to compute and graphics respectively
* When building the pipeline for the sphere rendering, the code leaves primitive restart enabled even though the sphere uses triangle list, not triangle strip.

This PR fixes these issues.  Additionally, it 

* removes one superfluous compute->compute barrier inside the iteration loop 
* switches the synchronization between queues to use semaphores instead of fences
  * as a consequence, the compute command buffers must be marked for simultaneous use.

### DIFF
--- a/examples/computecloth/computecloth.cpp
+++ b/examples/computecloth/computecloth.cpp
@@ -66,11 +66,14 @@ public:
 			vks::Buffer input;
 			vks::Buffer output;
 		} storageBuffers;
+        struct Semaphores {
+            VkSemaphore ready{ nullptr };
+            VkSemaphore complete{ nullptr };
+        } semaphores;
 		vks::Buffer uniformBuffer;
 		VkQueue queue;
 		VkCommandPool commandPool;
 		std::array<VkCommandBuffer,2> commandBuffers;
-		VkFence fence;
 		VkDescriptorSetLayout descriptorSetLayout;
 		std::array<VkDescriptorSet,2> descriptorSets;
 		VkPipelineLayout pipelineLayout;
@@ -133,7 +136,8 @@ public:
 		vkDestroyPipelineLayout(device, compute.pipelineLayout, nullptr);
 		vkDestroyDescriptorSetLayout(device, compute.descriptorSetLayout, nullptr);
 		vkDestroyPipeline(device, compute.pipeline, nullptr);
-		vkDestroyFence(device, compute.fence, nullptr);
+        vkDestroySemaphore(device, compute.semaphores.ready, nullptr);
+        vkDestroySemaphore(device, compute.semaphores.complete, nullptr);
 		vkDestroyCommandPool(device, compute.commandPool, nullptr);
 	}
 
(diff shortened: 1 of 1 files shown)


examples/computecloth/computecloth.cpp
```diff
@@ -694,23 +737,47 @@ class VulkanExample : public VulkanExampleBase
...
+	    computeSubmitInfo.commandBufferCount = 1;
+	    computeSubmitInfo.pCommandBuffers = &compute.commandBuffers[readSet];
+
+	    VK_CHECK_RESULT( vkQueueSubmit( compute.queue, 1, &computeSubmitInfo, VK_NULL_HANDLE) );
 
 		// Submit graphics commands
 		VulkanExampleBase::prepareFrame();
 
+        VkPipelineStageFlags waitDstStageMask[2] = {
+            submitPipelineStages, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT
```
cforfang: `BOTTOM_OF_PIPE_BIT` ->`VERTEX_INPUT_BIT` (wait on compute being done before reading the vertex data).

examples/computecloth/computecloth.cpp
```diff
@@ -694,23 +737,47 @@ class VulkanExample : public VulkanExampleBase
...
 	{
-	VkSubmitInfo computeSubmitInfo = vks::initializers::submitInfo();
-	computeSubmitInfo.commandBufferCount = 1;
-	computeSubmitInfo.pCommandBuffers = &compute.commandBuffers[readSet];
-
-	VK_CHECK_RESULT( vkQueueSubmit( compute.queue, 1, &computeSubmitInfo, compute.fence ) );
+        static bool firstDraw = true;
+        VkSubmitInfo computeSubmitInfo = vks::initializers::submitInfo();
+        // FIXME find a better way to do this (without using fences, which is much slower)
+        VkPipelineStageFlags computeWaitDstStageMask = VK_PIPELINE_STAGE_VERTEX_INPUT_BIT;
```
cforfang: `COMPUTE_SHADER_BIT`

examples/computecloth/computecloth.cpp
jherico: thanks, fixed

examples/computecloth/computecloth.cpp
jherico: thanks, fixed

SaschaWillems:
I'm already working on fixing synchronization and all related validation layer errors. See [this PR at the Khronos Vulkan Samples repository](https://github.com/KhronosGroup/Vulkan-Samples/issues/3).

Will check your changes as soon as I get some time to spare :)

examples/computecloth/computecloth.cpp
jherico: Does this not also imply that https://github.com/SaschaWillems/Vulkan/blob/master/examples/computecullandlod/computecullandlod.cpp#L806 is incorrect?  In other words the `VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT` isn't likely to ever be a correct destination wait stage for a command buffer containing only graphics pipeline operations?

cforfang: Yeah, agree that seems wrong as well.

SaschaWillems:
I'm currently lacking the time for digging into fixing all the compute shader synchronization issues.

So thanks for this PR, I really appreciate it :)

PR #651, opened 2020-01-08
Fix interface mismatch between VS and GS in deferredshadows

chrisforbes:
GLSLang determines the content of the builtin block based on the shading
language version. In the case of 420 vs 450, the difference is whether
the gl_CullDistance member is present.

Vulkan requires that the builtin interface block be the same on both
sides of a shader interface not involving the fragment shader, so this
leads to undefined behavior. This is caught by recent versions of the
validation layers.

Fix the issue by using GLSL 450 for both shaders in the pipeline.


SaschaWillems:

PR #652, opened 2020-01-10
Fix uninitialized variable used in vkCmdCopyBufferToImage().

NotCamelCase:
_offset_ used via ktxTexture_GetImageOffset was only initialized in debug mode.


SaschaWillems:
Thx for your PR.  Very much appreciated.

PR #658, opened 2020-01-25
Fix asset path for some examples

hhsaez:
I was trying to build _imgui_ and _pbrtexture_ examples on Mac OS and I've noticed they were using the ASSET_PATH macro instead of getAssetsPath() as other examples do. That made those examples fail when running because some assets where not found.

Please note that I'm still new to Vulkan and I'm not 100% sure the output from those examples is the expected one (after all, they're not listed in examples.h on Xcode project and ImGui does not seem to play nicely with retina displays). But since they run without crashes after these changes, maybe this PR may save you some time in the future.

### DIFF
--- a/examples/imgui/main.cpp
+++ b/examples/imgui/main.cpp
@@ -329,8 +329,8 @@ public:
 
 		pipelineCreateInfo.pVertexInputState = &vertexInputState;
 
-		shaderStages[0] = example->loadShader(ASSET_PATH "shaders/imgui/ui.vert.spv", VK_SHADER_STAGE_VERTEX_BIT);
-		shaderStages[1] = example->loadShader(ASSET_PATH "shaders/imgui/ui.frag.spv", VK_SHADER_STAGE_FRAGMENT_BIT);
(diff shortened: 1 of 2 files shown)


SaschaWillems:
Thanks for your PR. The ASSET_PATH was intentional and was meant as a replacement for getAssetPath, but I never got around making this consistent across all samples. I'll take a look at the changes and if they're okay I'll merge this.

Also note that I don't do any development on Apple devices, and the Mac OS part of this repo has been contributed by Brenwill (makers of MoltenVK), so I'm actually not able to test on MacOS. Will check in Windows, Linux and Android though.

hhsaez:
If ASSET_PATH is a replacement for getAssetPath, then I think I made things worse. Correct me if I'm wrong, but my understanding is that it was the other way around and getAssetPath was the newer mechanism, right?

SaschaWillems:
It's fine. I never got around using ASSET_PATH in all places, and it's also missing some functionality compared to getAssetPath. Will test this weekend and merge.

hhsaez:
Thanks!

PR #660, opened 2020-01-30
Fix the cmake_minimum_required for the libktx

weblucas:
I believe that 3.8 was a mistake. I am running 3.5, but all other CmakeFiles have 2.8 as minimum.


SaschaWillems:

PR #664, opened 2020-02-07
Clamp miplevel size to avoid zero width or height

gary-sweet:
I was seeing a zero height in certain circumstances for textures in scenerendering.
This change forces the minimum dimension to 1.


SaschaWillems:

PR #668, opened 2020-02-13
Perform sampling before branching to discard to avoid undefined behavior

kuhar:
This two fragment shaders that exhibit undefined behavior. In GLSL,
it is undefined to use implicit or explicit derivatives in
non-uniform control flow [1]. The SPIR-V produced by glslang contained
similar undefined behavior, which results in incorrect codegen
with the amdvlk's LLPC shader compiler, and produces visually
incorrect image.

This PR fixes this issue by hoisting the sampling so that it's
executed before discarding. It's a simple fix and may not be
the optimal way to resolve this issue.

[1] https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.60.pdf p. 137.


SaschaWillems:

PR #677, opened 2020-03-01
Update scenerendering.cpp

liuhongyi0101:
When merging indexes, we need to consider the number of vertices that have been merged

### DIFF
--- a/examples/scenerendering/scenerendering.cpp
+++ b/examples/scenerendering/scenerendering.cpp
@@ -312,7 +312,18 @@ private:
 			bool hasUV = aMesh->HasTextureCoords(0);
 			bool hasColor = aMesh->HasVertexColors(0);
 			bool hasNormals = aMesh->HasNormals();
-
+			
+			uint32_t currentVertexsCount = vertices.size();
+			// Indices
+			for (uint32_t f = 0; f < aMesh->mNumFaces; f++)
+			{
+				for (uint32_t j = 0; j < 3; j++)
+				{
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Can you elaborate? Does this crash with the current version of the sample?

liuhongyi0101:
the current version of the sample is right, but when I run your code with sponza , some of the models fail to be rendered.

liuhongyi0101:
> Can you elaborate? Does this crash with the current version of the sample?
the current version of the sample is right, but when I run your code with sponza , some of the models fail to be rendered.

liuhongyi0101:
> Can you elaborate? Does this crash with the current version of the sample?

the current version of the sample is right, but when I run your code with sponza , some of the models fail to be rendered.

liuhongyi0101:
> Can you elaborate? Does this crash with the current version of the sample?

Replaced with sponza model, some models cannot be rendered，Isn't this a bug?   I reference your project  [https://github.com/SaschaWillems/VulkanSponza ]，It is right 。

SaschaWillems:
Sorr  for the late reply. Just checked and Sponza didn't work for me either. I just pushed a fix that makes scenes other than Sibenik (like Sponza) to render properly witht that sample.

liuhongyi0101:
Fixes rendering for other models like sponza

> Sorr for the late reply. Just checked and Sponza didn't work for me either. I just pushed a fix that makes scenes other than Sibenik (like Sponza) to render properly witht that sample.

It's ok，The problem is solved

PR #704, opened 2020-04-30
Update WM_CLASS property on Linux using xcb

MattGuerrette:
This commit updates the WM_CLASS property to display a better
title than "Unknown" for each example in the dash and application menu.


SaschaWillems:
Sorry for the late reply. This is actually a good addition.

PR #720, opened 2020-05-25
xcb_connect always returns nonnull pointer, errors must be checked with xcb_connection_has_error

datenwolf:
`xcb_connect` always return nonnull pointers. The official documentation states

> Always returns a non-NULL pointer to a xcb_connection_t, even on failure. Callers need to use xcb_connection_has_error() to check for failure.

https://xcb.freedesktop.org/manual/group__XCB__Core__API.html#ga094470586356d1764e69c9a1882966c3


SaschaWillems:
Thx for your PR. I noticed that this also updates the ktx sub module. I recently had to fix this submodule to a tag to make it work with my repo (again), so that may be the cause for that. If you can revert the change to the sub module I'll merge this one.

SaschaWillems:
Thanks. Very much appreciated :)

PR #728, opened 2020-06-01
Some cleanup in base

hbina:
While reading through the codebase and learning about Vulkan, I noticed the following changes in base/

Edit: This PR is turning out to be more than just using STL.

Edit 2: Comprehensive list from my second commit:

```
1. Some member functions can be made `const`.
2. Avoid using C-style casting.
3. explicit constructors for constructor with 1 argument.
4. Use STL algorithms whenever possible. Also pre-allocate buffer before pushing.
5. Removed unreachable code (like `break` after `return` and `exit` after `return`).
6. Initialized many previously uninitialized variables.
7. Handle error return codes (mostly from freopen_s in Windows).
8. Avoid flushing with every print statement by replacing `std::endl` with "\n".
9. Some off-by-one error involving finding a random ratio algorithms. Mostly just avoid the warning by explicitly casting RAND_MAX to float.
```

### DIFF
--- a/base/benchmark.hpp
+++ b/base/benchmark.hpp
@@ -66,11 +66,11 @@ namespace vks
 					frameTimes.push_back(tDiff);
 					frameCount++;
 				};
-				std::cout << "Benchmark finished" << std::endl;
-				std::cout << "device : " << deviceProps.deviceName << " (driver version: " << deviceProps.driverVersion << ")" << std::endl;
-				std::cout << "runtime: " << (runtime / 1000.0) << std::endl;
-				std::cout << "frames : " << frameCount << std::endl;
-				std::cout << "fps    : " << frameCount / (runtime / 1000.0) << std::endl;
+				std::cout << "Benchmark finished" << "\n";
+				std::cout << "device : " << deviceProps.deviceName << " (driver version: " << deviceProps.driverVersion << ")" << "\n";
+				std::cout << "runtime: " << (runtime / 1000.0) << "\n";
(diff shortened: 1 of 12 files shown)


SaschaWillems:
I haven't had time to take a look at the PR yet, but could you please either disable travis CI on your fork or change the E-Mail? I'm getting notifications for all your failed builds.

SaschaWillems:
It would also be good if you could split up your PRs into smaller pieces. Some of the changes are okay, but other I'd like to decline. E.g. using std::find. I prefer raw loops due to better readability.

base/benchmark.hpp
```diff
@@ -36,8 +36,12 @@ namespace vks
...
 #if defined(_WIN32)
 			AttachConsole(ATTACH_PARENT_PROCESS);
-			freopen_s(&stream, "CONOUT$", "w+", stdout);
-			freopen_s(&stream, "CONOUT$", "w+", stderr);
+			if (freopen_s(&stream, "CONOUT$", "w+", stdout)) {
+				std::cerr << "Unable to redirect stdout to file" << "\n";
+			}
+			if (freopen_s(&stream, "CONOUT$", "w+", stderr)) {
+				std::cerr << "Unable to redirect stdout to file" << "\n";
+			}
```
eloj, line 44: "... stderr to file"

hbina:
Oops, sorry about that. I will get back on it after I did my homework. I have since turned this PR into a Draft. I hope it doesn't trigger the emails.

PR #731, opened 2020-06-05
GeometryShader/NormalDebugging: Using same uniform buffer for both pipelines.

ltqsoft:
I made the example simpler by using one uniform buffer for both the scene pipeline and normal debugging pipeline. And by adding the lineWidth into the uniform buffer data, normal vector can be made bigger or smaller with an ImGUI::sliderFloat.
I only tested it on my own machine, thus I don't know it worked on other machines or not.

OS: Windows 10 insider slow ring
Compiler/IDE: Visual Studio 2019 Community
Platform: x64 Debug


SaschaWillems:
Sorry for the late reply, but haven't had time to check this earlier. While this basically works, this PR misses a check for the `wideLines` physical device feature, which needs to be enabled and checked for support when using line widths > 1.0f.

PR #735, opened 2020-06-09
Fix: group INDEX_SHADOW_HIT 

michaeleggers:
Assign `shaderIndexShadowMiss` to `.generalShader` instead of `closestHitShader`, because:

closestHitShader must be either VK_SHADER_UNUSED_NV or a valid index into VkRayTracingPipelineCreateInfoNV::pStages referring to a shader of VK_SHADER_STAGE_CLOSEST_HIT_BIT_NV

( https://www.khronos.org/registry/vulkan/specs/1.1-extensions/html/vkspec.html#VUID-VkRayTracingShaderGroupCreateInfoNV-closestHitShader-02417 )

### DIFF
--- a/examples/nv_ray_tracing_shadows/nv_ray_tracing_shadows.cpp
+++ b/examples/nv_ray_tracing_shadows/nv_ray_tracing_shadows.cpp
@@ -595,9 +595,8 @@ public:
 		groups[INDEX_CLOSEST_HIT].closestHitShader = shaderIndexClosestHit;
 		// Shadow closest hit shader group
 		groups[INDEX_SHADOW_HIT].type = VK_RAY_TRACING_SHADER_GROUP_TYPE_TRIANGLES_HIT_GROUP_NV;
-		groups[INDEX_SHADOW_HIT].generalShader = VK_SHADER_UNUSED_NV;
 		// Reuse shadow miss shader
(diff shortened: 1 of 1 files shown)


manueme:
I think that this hit shader group is actually not being used on that example because the ray is traced with the "Skip Closest Hit Shader" flag. I'm sure you can even delete INDEX_SHADOW_HIT, remove the handle from the SBT, and change NUM_SHADER_GROUPS from 5 to 4.

michaeleggers:
> I think that this hit shader group is actually not being used on that example because the ray is traced with the "Skip Closest Hit Shader" flag. I'm sure you can even delete INDEX_SHADOW_HIT, remove the handle from the SBT, and change NUM_SHADER_GROUPS from 5 to 4.

Yes, thank you for pointing that out. I was simply in alert when the validation layers warned me about this. But as you say, the shader is simply skipped so it doesn't matter. But, maybe for correctness, the handle should be removed from the STB?

manueme:
(Y) I agree. Updating the PR removing that group (that also fixes the warning) may be the best call here.

SaschaWillems:
I have updated my ray tracing samples to use the KHR extension, and along with that reworked the whole shader group and shader binding table setup. This should also fix this issue.

PR #737, opened 2020-06-15
Fix offset for texture coordinates - vulkanscene example

manueme:
This example has no texture but it could be a problem if someone tries to use a different model


SaschaWillems:
Thanks for fixing this obvious mistake. Very much appreciated.

PR #740, opened 2020-06-19
shadowmappping: check if depth format supports LINEAR filtering

infapi00:
The test uses the format VK_FORMAT_D16_UNORM for the shadow map, and
it sets unconditionally VK_FILTER_LINEAR when using it.

But by spec, it is not mandatory that format to support
filtering. Explained here:
https://www.khronos.org/registry/vulkan/specs/1.2/html/chap32.html

table 51, VK_FORMAT_FEATURE_SAMPLED_IMAGE_FILTER_LINEAR_BIT.

This commit checks if that flag is present with that format to decide
between LINEAR (the default value) or NEAREST (if LINEAR is not
supported). Adds a auxiliar method on VulkanTools just in case it
could be useful for other demos.

This is not detected by the Validation Layers, but raise an assertion
with one of the development tools we use to implement the Mesa v3dv
driver (rpi4 vulkan driver).

### DIFF
--- a/base/VulkanTools.cpp
+++ b/base/VulkanTools.cpp
@@ -100,6 +100,21 @@ namespace vks
 			return false;
 		}
 
+		// Returns if a given format support LINEAR filtering
+		VkBool32 formatIsFilterable(VkPhysicalDevice physicalDevice, VkFormat format, VkImageTiling tiling)
+		{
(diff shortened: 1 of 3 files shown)


SaschaWillems:

PR #741, opened 2020-06-29
correction to yoffset char positions in distance field fonts example

Jarreddebeer:
While playing with the distanceFieldFont example I noticed the vertical alignment of characters seemed off. The text "Vulkan" does not seem affected but some charaters are, like "log":

rsz_screenshot_from_2020-06-29_20-57-56

After a lot of fiddling and trying to think upside down it seems this does the trick, but i'm not sure it's absolutely correct wrt baseline:

rsz_screenshot_from_2020-06-29_20-57-32

Please excuse such an insignificant pull request. I thought it worthwhile because it took me a good while to 'get right' and it otherwise confused me from the intent of the example.


SaschaWillems:
I don't see this as insignificant ;)

So thanks for your PR. Very much appreciated.

PR #755, opened 2020-08-20
Add Order Independent Transparency example

daemyung:


SaschaWillems:
Thank you very much for your PR. An OIT sample is something I always wanted to add, but never actually had the time to do so. So this is very much appreciated.

There are only a few small things that should be changed. I'll highlight them via separate comments.

examples/oit/oit.cpp
```diff
@@ -0,0 +1,683 @@
+/*
+* Vulkan Example - High dynamic range rendering
+*
+* Note: Requires the separate asset pack (see data/README.md)
+*
+* Copyright by Sascha Willems - www.saschawillems.de
+*
+* This code is licensed under the MIT license (MIT) (http://opensource.org/licenses/MIT)
+*/
```
SaschaWillems, line 9: This needs to be updated, as this refers to the HDR sample. Please also add your copyright below mine, so people can see that this sample was contributed b ypu.

examples/oit/oit.cpp
```diff
@@ -0,0 +1,683 @@
...
+
+#include "vulkanexamplebase.h"
+#include "VulkanglTFModel.h"
+
+#define ENABLE_VALIDATION true
```
SaschaWillems, line 14: This should be set to false by default

SaschaWillems (changes requested):

examples/oit/oit.cpp
daemyung: Oh it's my mistake... I have changed typos and added copyright.

examples/oit/oit.cpp
daemyung: Done.

daemyung:
> Thank you very much for your PR. An OIT sample is something I always wanted to add, but never actually had the time to do so. So this is very much appreciated.
> 
> There are only a few small things that should be changed. I'll highlight them via separate comments.

Thanks. I am happy to contribute. This implementation is very simple and performance is not good enough but

PR #756, opened 2020-08-25
Output warning to stderr if an extension is not supported

SirLynix:
Hi,

I was trying to run some raytracing examples and got an error box saying "Could not create Vulkan device: ERROR_EXTENSION_NOT_PRESENT". 

Since the raytracing examples enable a lot of extensions, I added a check before creating the instance / device to output to stderr what extensions are missing.

not related to this PR: is it normal that I'm missing VK_KHR_pipeline_library VK_KHR_deferred_host_operations and VK_KHR_ray_tracing with a RTX 2080 Ti and latest NVidia driver (452.06)?

### DIFF
--- a/base/VulkanDevice.cpp
+++ b/base/VulkanDevice.cpp
@@ -9,6 +9,7 @@
 */
 
 #include <VulkanDevice.h>
+#include <unordered_set>
 
 namespace vks
 {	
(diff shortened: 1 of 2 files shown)


SaschaWillems:
Thanks for your PR. I'll check the changes as soon as time permits.

As for the extensions missing: The ray tracing extensions are not finalized yet, and still considered beta. So they're not part of the main line driver, and only available in Nvidia's Vulkan developer driver.

PR #762, opened 2020-09-11
Run on MacOS

daemyung:


daemyung:
#if condition looks mess because Xcode project is already existed and we need to define functions and classes when the project is generated by CMake. I think supporting both barebone Xcode and CMake is fine because some developers want to set up Vulkan on thier barebone Xcode project. I can't run all examples because the number of examples is too many. I have checked some examples and examples are work except geometry shader examples. It's not supported on Apple platform.

daemyung:
Also examples using image atomic operation don't work. It looks like MoltenVK issues. After you merge this patch, I will report to MoltenVK.

SaschaWillems:

PR #770, opened 2020-10-06
Support macOS with clang.

prideout:
This fixes the build errors that look like:

    source file is not valid UTF-8

This was due to CMake adding object files to the command line after the `-x` argument, which caused them to be interpreted as Objective C++ source code.


SaschaWillems:

PR #782, opened 2020-11-26
📝 use gradlew

m0ppers:
this should ease building on Android. AFAIK (neither java nor android expert) and if I understand https://docs.gradle.org/current/userguide/gradle_wrapper.html correctly this should be best practice.

At least this is what I am doing in my projects.


SaschaWillems:

Issue #50, opened 2016-02-28
Rendering to cubemap face directly in shadowmappingomni demo


httpdigest:
I would normally haved asked this on the forums.khronos.org directly, but this question also relates to the shadowmappingomni demo, so I ask it here:

I saw that this demo creates a single 2D framebuffer image, renders each of the six views of the cubemap onto it and does each an image copy of the framebuffer content to the cubemap faces' images.

Would it be possible to use the cubemap face's image/view directly as framebuffer image and directly render to it? Of course, one would have to do the memory barrier to switch between renderable and sourceable, but would it not avoid the memcpy?

Sorry to take your time on this. I don't require an elaborate explanation of why or not this would be possible, a simple "yes" or "no" suffices for me to do research on my own. :)
But I reckon that this would be possible. But likely then with six framebuffers or a geometry shader that directs to the right layer of a layered framebuffer attachment/image?

Thanks!

httpdigest:
I did some digging around in the specs and navigating the Vk structs, and it seems rendering to each of the six cubemap faces is doable in a single renderpass, by using six render subpasses each having a different VkSubpassDescription.pColorAttachments and .pDepthStencilAttachment which each point to a VkAttachmentReference with its own index/binding point.
Then when creating the framebuffer with vkCreateFramebuffer, assign the attachment binding points in VkFramebufferCreateInfo to each of the six different image views.

SaschaWillems:
That sounds reasonable. Haven't tried it myself, but it should work. And if it works I'd love to hear back :)

Layered rendering was something I considered. Using he cube map faces as frame buffers should work too (I guess) if you set the correct flags. 

But layered rendering (and duplicating in the geometry shader) may not be worth it and depending on hardware it may even be slower than just doing the separate rendering passes.

httpdigest:
Okay, thanks. I'd close this issue now and have a go with the separate subpasses rendering to the cubemap images, and let you know how that went.

ChristophLGDV:
I have implemented a sample using the geometry shader, that renders and samples the same texture (combining your deferedshadow and shadowmapping and omnishadow).
The preformance drops significantly when invoking additional Geometry (fixed budget i guess, 2 or 6 does not make much difference.) (970GTX)

I don't think that I missed something but maybe you can have a look. 
I will try on pascal tomorrow.

ChristophLGDV:
Pascal is better than Maxwell, I get better performance compared to the copy solution. 

[...]

PR #783, opened 2020-11-28
Omnidirectional Shadow Mapping example - Render depth to cubemap faces directly

michead:
Render depth to cubemap faces directly, while avoiding copying 6 images per frame, one per cubemap face.
This approach was discussed in issue #50. 
The resulting code is terser, hence easier to read, and more performant (at least on my machine).
The changes have been tested on both Windows and macOS (via MoltenVK).


SaschaWillems:
Sorry for the very very late reply. Doing some end of the year cleanup and realized this was still open. Just merged this one. Not only is the code easier to read, but it's also faster by a good margin.

So thank you very much for this PR!

PR #785, opened 2020-12-04
Don't use pragma once in .cpp file

kuhar:
Fix a compiler warning.


SaschaWillems:

PR #797, opened 2021-01-07
Fix long options for width and height

caramelli:
Two dashes should probably be used as for the other long options.

Nicolas Caramelli


SaschaWillems:

PR #802, opened 2021-01-18
Change scope of validationExt and validationLayerName variables

jeromenxp:
Pointer to a variable declared into a block statement is stored and used outside of the block by vkCreateInstance. This can lead to segmentation fault.
Move the declaration of validationExt/validationLayerName outside of the block to fix the crash.


SaschaWillems:

PR #807, opened 2021-02-15
Fix compile warning and crash in SSAO test.

per-mathisen-arm:
emptyTexture was not initialized, so when destroying it the destroy Vulkan functions were called with bogus pointers.


SaschaWillems:

PR #811, opened 2021-02-23
Invalid VkSpecializationInfo passed into vkCreateRayTracingPipelinesKHR

mdinkov:
The test declares VkSpecializationInfo in a localized scope and calls vkCreateRayTracingPipelinesKHR outside of that scope. Thus, by the time vkCreateRayTracingPipelinesKHR is called, VkSpecializationInfo has been destroyed.


SaschaWillems:

PR #812, opened 2021-02-24
Remove acceleration structure host commands usage

tangmi:
Removes usage of `vkBuildAccelerationStructuresKHR` from the ray tracing examples.

- According to http://vulkan.gpuinfo.org/, there are currently no drivers that support host commands, so this code was not tested or testable.
- The comment added is from [the ray tracing sample at KhronosGroup/Vulkan-Samples](https://github.com/KhronosGroup/Vulkan-Samples/blob/55cebd9e7cc4153a3a7b3a45d42274c0e2a17815/samples/extensions/raytracing_basic/raytracing_basic.cpp#L303-L304).


SaschaWillems:
Sorry for the long delay. Somehow totally missed this. Your PR is very much appreciated!

PR #813, opened 2021-03-01
Variable types corrected

LeonMike:
Not sure if this issue is because of glm version, but I think this changes could be helpful to avoid some castings.


SaschaWillems:
Did you check if the samples work as intended with the glm version from the repo? Some of the odd looking casts, esp. those related to quaternions, were necessary to make things work.

LeonMike:
> Did you check if the samples work as intended with the glm version from the repo? Some of the odd looking casts, esp. those related to quaternions, were necessary to make things work.

Not really, I haven't had enough time to test the app. The glm I used was the current available version in OpenSUSE Leap 15.2. I'll give it a try and will come to you with the answer

LeonMike:
@SaschaWillems everything looks fine after the changes using the glm version in the repo and the changes included into this pull-request.

PR #829, opened 2021-04-29
Missing fences in compute scenarios before UBO update on CPU

adamjer:


SaschaWillems:
Thanks for your PR, it's very much appreciated. I'm currently in the process of doing a basic rewrite of how synchronization works in my samples and will incorporate your changes into that at a later point.

PR #830, opened 2021-04-29
Add a `Running` section to the readme that provides info about multi-gpu setups

expenses:
It wasn't clear to me that there were command line options in the examples, or that the gpu that is used could be selected. I have an integrated internal gpu and a discrete external gpu, but for all the examples (most irritatingly the ray tracing ones) the internal gpu is selected by default.

I'm happy to change these changes in any way you see fit.


SaschaWillems:
That looks great. Thank you very much for your PR.

PR #855, opened 2021-07-15
Add a build script

xykivo:
The build script will

1. Update submodules
2. Download assets
3. Generate build files
4. Build

This change also modifies the download assets script

1. The download code is in a function, and is called from the script main. This is done to enable calling this script from the build script
2. Skip downloading the dependencies if they were already downloaded

For this script only supports desktop builds.

### DIFF
--- a/build.py
+++ b/build.py
@@ -1,0 +1,132 @@
+# Copyright (c) 2021, Dror Smolarsky
+# All rights reserved.
+#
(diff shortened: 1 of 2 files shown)


SaschaWillems:
I'll change the assets to a submodule in the future, meaning no more download scripts required.

PR #859, opened 2021-07-29
Fix rayquery demo to not require full raytracing support.

per-mathisen-arm:


SaschaWillems:

PR #860, opened 2021-07-29
Add option to only render a given number of frames.

per-mathisen-arm:
This is useful on very slow targets, such as GPU emulators.


SaschaWillems:

PR #862, opened 2021-08-03
Fixed the first frame render error.

ShoufengYang:
The compute commands should be submit before the graphic command. If not, there is a logical error that cause the first frame render error.  Since the graphic rendering need to sample the compute shader 's output image.


SaschaWillems:
Thanks for your PR. Please give me some time to review this.

Issue #872, opened 2021-08-31
dynamicuniformbuffer example validation error


kyamant:
In the function `buildCommandBuffers()` the call `vkCmdBindDescriptorSets(drawCmdBuffers[i], VK_PIPELINE_BIND_POINT_GRAPHICS, pipelineLayout, 0, 1, &descriptorSet, 1, &dynamicOffset);`, starting with iteration j=1 (j =0 has dynamicOffset = 0 and that does not represent a problem)  results in:

`ERROR: [-1587959965][VUID-vkCmdBindDescriptorSets-pDescriptorSets-01979] : Validation Error: [ VUID-vkCmdBindDescriptorSets-pDescriptorSets-01979 ] Object 0: handle = 0x2a38df0bfe8, type = VK_OBJECT_TYPE_COMMAND_BUFFER; Object 1: handle = 0x4295ab0000000035, type = VK_OBJECT_TYPE_DESCRIPTOR_SET; Object 2: handle = 0xa43473000000002d, type = VK_OBJECT_TYPE_BUFFER; | MessageID = 0xa159a763 | vkCmdBindDescriptorSets(): pDynamicOffsets[0] is 0x100, but must be zero since the buffer descriptor's range is VK_WHOLE_SIZE in descriptorSet #0 binding #1 descriptor[0]. The Vulkan spec states: For each dynamic uniform or storage buffer binding in pDescriptorSets, the sum of the effective offset, as defined above, and the range of the binding must be less than or equal to the size of the buffer (https://vulkan.lunarg.com/doc/view/1.2.182.0/windows/1.2-extensions/vkspec.html#VUID-vkCmdBindDescriptorSets-pDescriptorSets-01979)`

cforfang:
Hi @sjfricke, I noticed you changed the validation code for this in February (see https://github.com/KhronosGroup/Vulkan-ValidationLayers/commit/11b31217ff31d135baab22ab3e2d04d63ffd850b). The new code looks a bit wrong for the case of VK_WHOLE_SIZE.

Old (compared dynamic + base offset vs buffer size):
```
 if (VK_WHOLE_SIZE == range) {
    if ((dyn_offset + desc_offset) > buffer_size) {
       // ERROR
```
New (error out for offsets greater than zero when using VK_WHOLE_SIZE):
```
 // Validate offset didn't go over buffer
if ((bound_range == VK_WHOLE_SIZE) && (offset > 0)) {
   // ERROR
```

[...]

PR #873, opened 2021-08-31
Fix bound range in dynamicuniformbuffer sample

cforfang:
Untested fix for #872 -- I only checked the code builds.


cforfang:
Confirmed this fixes the validation layer error and rendering output looks good. Tested using Vulkan SDK 1.2.176.1 and VK_LAYER_KHRONOS_validation only.

SaschaWillems:

Issue #814, opened 2021-03-08
Sparse texture example leaves mips > 0 in UNDEFINED layout


baldurk:
The [texturesparseresidency](https://github.com/SaschaWillems/Vulkan/tree/master/examples/texturesparseresidency) example only seems to barrier/transition image layout for mip 0. This produces validation layer errors because the other mips are in UNDEFINED:

> ERROR: [1303270965][UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout] : Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidImageLayout ] Object 0: handle = 0x52f21b8, type = VK_OBJECT_TYPE_COMMAND_BUFFER; | MessageID = 0x4dae5635 | Submitted command buffer expects VkImage 0xd20a5a000000003f[] (subresource: aspectMask 0x1 array layer 0, mip level 1) to be in layout VK_IMAGE_LAYOUT_SHADER_READ_ONLY_OPTIMAL--instead, current layout is VK_IMAGE_LAYOUT_UNDEFINED.

And similarly for mips 2-13.

The overload of `vks::tools::setImageLayout` that is called within the sample only transitions the first mip in the first slice (I assumed a shortcut for the common case of a render target with no mips/slices). I locally switched to store the subresource range and use the other overload, but there may be a better fix you prefer especially if you don't want to transition the whole image every time:

```patch
diff --git a/examples/texturesparseresidency/texturesparseresidency.cpp b/examples/texturesparseresidency/texturesparseresidency.cpp
index 1c7a348..a47bfe1 100644
--- a/examples/texturesparseresidency/texturesparseresidency.cpp
+++ b/examples/texturesparseresidency/texturesparseresidency.cpp
@@ -187,6 +187,8 @@ void VulkanExample::prepareSparseTexture(uint32_t width, uint32_t height, uint32
 	texture.layerCount = layerCount;
 	texture.format = format;
 
+	texture.subRange = { VK_IMAGE_ASPECT_COLOR_BIT, 0, texture.mipLevels, 0, 1 };
+
 	// Get device properties for the requested texture format
 	VkFormatProperties formatProperties;
 	vkGetPhysicalDeviceFormatProperties(physicalDevice, format, &formatProperties);
@@ -236,7 +238,7 @@ void VulkanExample::prepareSparseTexture(uint32_t width, uint32_t height, uint32
[...]

PR #882, opened 2021-09-28
Fix texturesparseresidency memory and image layout bugs

neither-nor:
In the example, when flushing pages, the memory was freed before unbinding, this caused the page to become unable to be bound  again on my machine, changed it so we free after unbinding.
Also made some minor change so we don't rebind every page every time.
#814 Also makes sense to me so I included the change @baldurk proposed.


SaschaWillems:
Thank you very much for your PR. Very much appreciated. Will review this soon and merge.

SaschaWillems:
@kyamant: I have pushed a few changes to silence the validation layers.

kyamant:
Thank you for this repository. It is extremely helpful in learning Vulkan,

PR #907, opened 2021-12-03
Fixes memory leak due to ImGui context not being destroyed.

rajnesh-kanwal:
Found while testing using Valgrind with triangle demo.

### DIFF
--- a/base/VulkanUIOverlay.cpp
+++ b/base/VulkanUIOverlay.cpp
@@ -50,7 +50,9 @@ namespace vks
 		io.FontGlobalScale = scale;
 	}
 
-	UIOverlay::~UIOverlay()	{ }
+	UIOverlay::~UIOverlay()	{
+		ImGui::DestroyContext();
+	}
 
 	/** Prepare all vulkan resources required to render the UI overlay */
 	void UIOverlay::prepareResources()
@@ -397,7 +399,6 @@ namespace vks
 
 	void UIOverlay::freeResources()
 	{
-		ImGui::DestroyContext();
 		vertexBuffer.destroy();
 		indexBuffer.destroy();
 		vkDestroyImageView(device->logicalDevice, fontView, nullptr);


rajnesh-kanwal:
@SaschaWillems, Can you add the appropriate reviewers to this PR? It's my first contribution so I am unable to do that.
Thanks.

SaschaWillems:
Thanks for your PR.

The UIOverlay class already frees the ImGUI context, see https://github.com/SaschaWillems/Vulkan/blob/master/base/VulkanUIOverlay.cpp#L400

I guess this is not properly called in the triangle sample, so that sample needs to be fixed. I'll look into that.

rajnesh-kanwal:
> Thanks for your PR.
> 
> The UIOverlay class already frees the ImGUI context, see https://github.com/SaschaWillems/Vulkan/blob/master/base/VulkanUIOverlay.cpp#L400
> 
> I guess this is not properly called in the triangle sample, so that sample needs to be fixed. I'll look into that.

Okay, I see what you are saying but still, there is a problem. `UIOverlay::freeResources()` is called from
https://github.com/SaschaWillems/Vulkan/blob/313ac10de4a765997ddf5202c599e4a0ca32c8ca/base/vulkanexamplebase.cpp#L853 but that depends on `settings.overlay` being true, but the UIOverlay object creation doesn't depend on it and the constructor is called regardless of this being set or not. Also, from the design perspective, I think the `ImGui::DestroyContext()` should be called from destructor given its being created in the constructor. 

I have updated the patch to remove the `ImGui::DestroyContext()` call from `UIOverlay::freeResources()`.

Also, I think it would be better to have the UIOverlay class object pointer rather than the object itself in `VulkanExampleBase` https://github.com/SaschaWillems/Vulkan/blob/313ac10de4a765997ddf5202c599e4a0ca32c8ca/base/vulkanexamplebase.h#L183 and the object should be created if `settings.overlay` is set. Right now it's created implicitly and destroyed implicitly with `VulkanExampleBase` class.

PR #923, opened 2022-01-21
fix example oit build for Android failed

robotchaoX:
[Why]
Example oit build for Android failed with error _"examples/oit/oit.cpp:552:4: error: use of undeclared identifier 'vkCmdFillBuffer'"_.
For Android, it build with parameter "-DVK_NO_PROTOTYPES". So, it will not use the vkCmdFillBuffer declaration under external/vulkan/vulkan_core.h.

[How]
Declare vkCmdFillBuffer in file base/VulkanAndroid.hpp&cpp as others did.


robotchaoX:
'vkCmdFillBuffer' was introduced to "examples/oit/oit.cpp" from commit fce1c93266.

SaschaWillems:

PR #932, opened 2022-03-06
Don't set VK_IMAGE_ASPECT_STENCIL_BIT bit if format is depth only

rajnesh-kanwal:
The fix is already present in VulkanExampleBase::setupDepthStencil. The problem with the current code is when creating a depth image with `VK_FORMAT_D32_SFLOAT` format, the image->aspect is set to `VK_IMAGE_ASPECT_DEPTH_BIT` only based on the format. In other terms image will only support `VK_IMAGE_ASPECT_DEPTH_BIT` aspect if format is  `VK_FORMAT_D32_SFLOAT`. While creating image_view it's not right to pass aspects that are not supported by the image, the view is being created from. 

The change checks if the format has the stencil part in it and only set `VK_IMAGE_ASPECT_STENCIL_BIT` bit then.

> aspectMask is a bitmask of [VkImageAspectFlagBits](https://www.khronos.org/registry/vulkan/specs/1.0-extensions/html/vkspec.html#VkImageAspectFlagBits) specifying which aspect(s) of the image are included in the view.

### DIFF
--- a/examples/triangle/triangle.cpp
+++ b/examples/triangle/triangle.cpp
@@ -661,7 +661,11 @@ public:
 		depthStencilView.viewType = VK_IMAGE_VIEW_TYPE_2D;
 		depthStencilView.format = depthFormat;
 		depthStencilView.subresourceRange = {};
-		depthStencilView.subresourceRange.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT;
+		depthStencilView.subresourceRange.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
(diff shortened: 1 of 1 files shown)


rajnesh-kanwal:
@SaschaWillems, I am not able to add any reviewers to the PR. Can you please add the appropriate reviewers? 
Thanks.

rajnesh-kanwal:
@SaschaWillems, sorry to ping you again. I would much appreciate it if you can please look into this PR, its quite a trivial change and shouldn't take long I hope. This is actually blocking [pvr: Add a Vulkan driver for Imagination Technologies PowerVR Rogue GPUs ](https://gitlab.freedesktop.org/mesa/mesa/-/merge_requests/15243).
Your help is much appreciated.

Thanks.

SaschaWillems:
Sorry for the late reply. Changes look good, will merge this ASAP.

PR #934, opened 2022-03-21
Fixed loadAnimation function memory leak bug.

jonnxie:
loadAnimation function has three memory leaks


SaschaWillems:

PR #936, opened 2022-03-22
Fixed GLTF memory leak

jonnxie:
The loadNode function has three memory leaks.Destruct function of Mesh class and Model class has memory leak.


SaschaWillems:
Thanks again for your PR. Very much appreciated!

PR #937, opened 2022-04-07
Ray Query example: shader optimization

JoseEmilio-ARM:
Calling rayQueryProceedEXT in a loop can prevent certain shader optimizations. By setting the TerminateOnFirstHit and SkipAABB flags, there is no scenario where rayQueryProceedEXT will return true, hence the loop can be removed. 

This way the implementation has a guarantee that the traversal can be completed without returning control to the shader, which improves performance.


SaschaWillems:

PR #959, opened 2022-06-30
Fix various cases of image views with incorrect stencil aspect

itoral:
In cases where the selected depth/stencil format is depth-only the view should  only contain the depth aspect. This issue can make Mesa drivers assert on debug builds, since they check this when views are created.


itoral:
FWIW, this is also detected by a recent enough version of validation layers:

`UNASSIGNED-CoreValidation-DrawState-InvalidImageAspect(ERROR / SPEC): msgNum: -1863356067 - Validation Error: [ UNASSIGNED-CoreValidation-DrawState-InvalidImageAspect ] Object 0: handle = 0x130160000000005c, type = VK_OBJECT_TYPE_IMAGE; | MessageID = 0x90ef715d | vkCreateImageView(): Using format (VK_FORMAT_D32_SFLOAT) with aspect flags (VK_IMAGE_ASPECT_DEPTH_BIT|VK_IMAGE_ASPECT_STENCIL_BIT) but depth-only image formats can have only the VK_IMAGE_ASPECT_DEPTH_BIT set.`

SaschaWillems:

PR #960, opened 2022-06-30
Don't require D32S8 format for deferred shadows sample

itoral:
The format is not mandatory by the spec and the sample doesn't
require stencil, so just choose a supported depth/stencil format
instead to allow this to run on more platforms.


SaschaWillems:

PR #965, opened 2022-07-13
Update macOS/iOS implementations, fix queue family transfers, implement window resizing for all examples

SRSaunders:
This is a fairly sizeable pull request that updates the macOS/iOS implementation for on-par functionality vs. Windows and Linux, but is still subject to Vulkan API limitations imposed by MoltenVK (e.g. no geometry shaders,  no raytracing, etc).  I probably should have broken this contribution into multiple smaller pull requests, but feature and regression testing on macOS, iOS (both sim and hardware), Windows, and Linux proved more efficient this way.

In addition, I have made some generic changes that hopefully helps for all platforms:
1. implemented window resizing `viewChanged()` for all examples, including multiview and multisampling
2. fixed queue family transfer operations (graphics <-> compute) in the compute* examples so they follow Vulkan sample code and no longer throw validation errors
3. updated `getQueueFamilyIndex()` to support `VkQueueFlags` vs. `VkQueueFlagBits` for the possibility of requesting a queue with multiple flag bits set - while not strictly necessary, I made this change during testing and is a generalization
4. fixed most Vulkan validation errors (with the exception of window resizing on linux where the validation layer's view of swapchain dimensions can get out of sync with the actual runtime and throw spurious false positives)
5. implemented the ability to change the swapchain image count on window resize events - while this was motivated by a macOS vsync issue for windowed vs fullscreen modes, it was an interesting problem and the solution is generic
6. enabled ImGui scaling for macOS retina displays - this could possibly be leveraged for other HiDPI situations
7. updated the computecloth example to use an animation rate that is independent of frame rate
8. updated the ImGui example to display the Vulkan API level + driver information and also defined initial window placement/sizing where possible

The macOS/iOS specific changes are:
1. updated the vulkanExample project (cmake -G "Xcode" or VK_EXAMPLE_XCODE_GENERATED) to support max frame rate rendering by default (the command line -vs option now works to optionally enable vsync if requested), the macOS and iOS xcode example projects remain vsync only (DisplayLink-driven).  Note: also fixes benchmarking (-b) on macOS.
2. updated the macOS xcode example project to properly handle and respond to keyboard and mouse events
3. updated the iOS xcode example project to properly handle and respond to touch, tap, pan, and pinch/zoom gestures
4. updated the iOS xcode example project to support the Xcode iOS simulator
5. implemented Vulkan / memory cleanup on shutdown for both vulkanExample and xcode example projects
6. fixed OpenMP integration for macOS and iOS - this accelerates texture regeneration in the texture3d example
7. vulkanExample windows now launch in front of Xcode vs behind it when debugging - an annoying issue fixed
8. updated the `VK_KHR_portability_subset` and `VK_KHR_portability_enumeration` implementation to perform runtime checks in case the project is linked to MoltenVK directly vs. to the Vulkan loader - cannot assume `VK_KHR_portability_enumeration` is available unless it appears at runtime (note this extension is Vulkan loader only)
9. fixed the glTF examples so they build and link properly for both vulkanExample and xcode example projects
10. fixed the triangle example so validation errors are eliminated and window resizing is possible on macOS (changes done using #defines so that bare code remains in place for other platforms, including iOS)
11. worked around a known MoltenVK vsync issue where frame rate appears to be double the monitor refresh rate - workaround is to manipulate swapchain image count to force image acquire frame rate to match display vsync frame rate
12. allow mouse click-through in vulkanExample project to capture first mouse click during long load-time examples (e.g. glTFloading example)

All these changes have been tested on macOS Big Sur, iOS 15.3.1 on an iPhone 8, iOS Sim for iPhone 13, Windows 10, and Linux Manjaro.

While there are a lot of changes here, hopefully they are useful to others and could help the project.


SRSaunders:
Thanks for looking at the pull request.  I started off with the intention of making only Apple-specific changes since the macOS/iOS portions of the project had not been updated for a while.  However, once I got into it I extended the scope to include validation and sync issues to hopefully improve things cross-platform.  Unfortunately I was not aware of your work in those areas and in retrospect perhaps this was a mistake on my part.  In any case, hopefully my proposed changes could be a point of comparison for your own modernization efforts.  At least for the graphics <-> compute queue sync issues, most of my changes were concentrated within commit b2f501dc98c967ec5f49d2e47d4f4975753b2a48.

As an aside, this pull request also fixed benchmarking on macOS (I forgot to include that in my notes above).  To that end I just added a small change updating the benchmark-all.py script to update the example list and recognize macOS.

After reviewing the changes, if you need any assistance separating the macOS/iOS - specific parts from the generic work I would be happy to help.  I could also do some post-merge testing for those platforms if you need it.  Let me know.

SaschaWillems:
Took a first deeper look at this. And while I can't judge the Apple side of things, Windows, Linux and Android look fine to me. Did a quick test with several samples on Android and they worked as expected.

I'll do a more detailed review on this PR very soon, but I think this is good to go :)

SRSaunders:
Glad to hear that Android seems to be working fine.  I don’t have that platform to test on.

Would you be able to check the imgui and gltfscenerendering examples on Android and post screen caps?  I would like to ensure UI scaling and (x,y) input scaling (for widget interaction) is working properly on that platform.  I want to make sure I didn’t break anything when implementing UI scaling for macOS retina displays.

SaschaWillems:
The samples mentioned look exactly the same on Android as on the master branch. So it seems that no further changes are required.

SRSaunders:
Good news - thanks for letting me know re Android scaling.

FYI - I found an additional issue with the master branch: the F1 gui overlay toggle does not work across most examples and platforms, in some cases generating validation errors.  I have a set of changes that will correct this.  Unless you request otherwise, I will wait until you are finished reviewing this pull request and hopefully merge it as is.  Once this is done I can submit another PR with the changes.  Would that approach work for you?

SaschaWillems:
A question on resizing things in case swap chain image count changes: Is that a thing on iOS or Mac? I intentionally did not implement that (see e.g. the multiview sample) cause I thought this should never happen during runtime of a sample. But if that can happen on one of these platforms, I'm fine with those changes.

examples/vertexattributes/vertexattributes.h
```diff
@@ -9,7 +9,7 @@
 #define TINYGLTF_IMPLEMENTATION
 #define STB_IMAGE_IMPLEMENTATION
 #define TINYGLTF_NO_STB_IMAGE_WRITE
-#define TINYGLTF_NO_STB_IMAGE
```
SaschaWillems, line 12: Can you elaborate on this? It works fine for me with the define in place. Does this cause problems on iOS/MacOS?

examples/computeraytracing/computeraytracing.cpp
```diff
@@ -94,6 +94,9 @@ class VulkanExample : public VulkanExampleBase
 
 	~VulkanExample()
 	{
+		// SRS - Ensure all operations on the device have finished before destroying resources
+		vkDeviceWaitIdle(device);
```
SaschaWillems, line 98: This shouldn't be necessary. The base class already does a `vkDeviceWaitIdle` at the end of the render loop.

examples/triangle/triangle.cpp
```diff
@@ -126,6 +126,9 @@ class VulkanExample : public VulkanExampleBase
 
 	~VulkanExample()
 	{
+		// SRS - Ensure all operations on the device have finished before destroying resources
+		vkDeviceWaitIdle(device);
```
SaschaWillems, line 130: This shouldn't be necessary. The base class already does a `vkDeviceWaitIdle` at the end of the render loop.

examples/CMakeLists.txt
```diff
@@ -39,14 +59,15 @@ function(buildExample EXAMPLE_NAME)
 	endif(WIN32)
 
 	set_target_properties(${EXAMPLE_NAME} PROPERTIES VS_DEBUGGER_WORKING_DIRECTORY ${CMAKE_SOURCE_DIR}/bin)
-	if(OpenMP_CXX_FOUND)
-		target_compile_options(${EXAMPLE_NAME} PRIVATE ${OpenMP_CXX_FLAGS})
+	#if(OpenMP_CXX_FOUND)
+		# SRS - target_compile_options is not needed when using target_link_libraries(<target> OpenMP::OpenMP_CXX) syntax below
```
SaschaWillems, line 63: I'm fine with just removing these lines then if they're not necessary.

SaschaWillems:
I went through all the changes and only noticed very minor things, see comments above.

examples/CMakeLists.txt
SRSaunders: > A question on resizing things in case swap chain image count changes: Is that a thing on iOS or Mac? I intentionally did not implement that (see e.g. the multiview sample) cause I thought this should never happen during runtime of a sample. But if that can happen on one of these platforms, I'm fine with those changes.

There is a difference with MoltenVK regarding where the code blocks when in vsync mode.  On most other platforms Vulkan code blocks on vkAcquireNextImageKHR() whereas with MoltenVK it blocks on vkQueueSubmit().  This can cause issues with apparent frame rate when in vsync mode.  If there are excess swap chain images available, then the frames per second can be double what you are expecting, but the overall additive frame period of the two frames together will be correct.  Nonetheless this causes havoc with frame rate measurement code in applications.  To work around this issue for the Vulkan Samples project, I added code to adjust the macOS swap chain image count based on whether in fullscreen mode (minImageCount + 1) or in windowed mode (minImageCount).  This works around the problem and results in correct vsync frame rates.  When outside of vsync mode, this adjustment is not required.

If possible I would prefer you leave the resize code in place to handle swap chain image count changes.  While the original issue is macOS-specific, it is a general solution and works on other platforms too.

examples/vertexattributes/vertexattributes.h
SRSaunders: Good comment and catch here.  I originally made this change (also in gltfscenerendering.h) based on build failures on macOS and iOS.  However, I later made a change to how VulkanglTFModel.cpp is included in the build config for the xcode/examples project (now conditionally included in examples.h when tiny_gltf.h is not otherwise included or customized by an example).  This build config change seems to make the change here unnecessary.  It's been a while since I looked at this area of the code, but now after uncommenting `#define TINYGLTF_NO_STB_IMAGE` I can no longer duplicate build failures for the vertexattributes and gltfscenerendering examples on macOS and iOS.  I guess you can revert this change here and in gltfscenerendering.h

examples/computeraytracing/computeraytracing.cpp
SRSaunders: This one is trickier.  On macOS interactive mode exit (not benchmarking exit) does not run through the end of renderLoop().  For the cross-platform `cmake -G Xcode` (VK_EXAMPLE_XCODE_GENERATED) project, the render loop is taken over by `[NSApp run]`.  Interactive termination is handled by the `applicationWillTerminate()` event handler and the exit code of renderLoop() never runs.  The only time it will run is when benchmarking on macOS and `[NSApp run]` is never started.  Your comment points out that I should have placed a `vkDeviceWaitIdle(device)` in the macOS termination event handler instead of in several example cases.  I should be able to fix this quite easily, reverting the example cases.  I can push a small change to the PR but I should test it first.

examples/triangle/triangle.cpp
SRSaunders: See comment above - same issue and same solution.

examples/CMakeLists.txt
SRSaunders: > I'm fine with just removing these lines then if they're not necessary.

Let me look into this.  It's been a while and I would like to review before answering.  Stay tuned.

SRSaunders: I have a new commit prepared with the above code review changes but I would like to test before submitting.  I should be able to push it soon and revise the PR based on your feedback.

SRSaunders:
> I went through all the changes and only noticed very minor things, see comments above.

Thanks very much for your feedback.  I have a new commit prepared with the above code review changes but I would like to test before submitting. I should be able to push it soon and revise the PR.

SaschaWillems:
Changes look good to me. Will merge this now :)

SRSaunders:
Thanks for merging. I appreciate your effort in reviewing this and including in the main branch.

I am away for a few days but will submit a new PR with the F1 UI overlay toggle fixes when I return.

luck-bear:
您好，邮件已收到，谢谢。会尽快给回复。

CMakeLists.txt
```diff
@@ -130,11 +130,11 @@ endif()
 IF(MSVC)
 	SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} /EHsc")
 ELSEIF(APPLE)
-	if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
-		SET(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -fobjc-arc -ObjC++")
-	ELSE()
+	#if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
```
awesomekosm, line 133: commenting this out breaks fixed issue https://github.com/SaschaWillems/Vulkan/pull/770
revert?

luck-bear:
您好，邮件已收到，谢谢。会尽快给回复。

CMakeLists.txt
SRSaunders: The issue here is the kind of cmake generator used for macOS builds.  The comment-out was to meant to support Xcode builds.  However, you are right that it appears to break macOS builds when using cmake's make or ninja command-line generators.  I will fix this by testing for CMAKE_GENERATOR and selecting the correct objective c++ option.  Note the previous test for "Clang" did not make sense since it has been the default macOS compiler for ages and is used for both Xcode and command line builds.  I will link a pull request here when ready.

SRSaunders: Changes implemented in PR #1117.  Please let me know if this fixes your issue.

PR #974, opened 2022-08-02
UI Overlay: Fix F1 on/off toggle for all examples, Implement overlay for dynamicrendering example

SRSaunders:
This is a small pull request that fixes three issues:

1. Fixes F1 on/off toggle for the ImGui UI overlay panel for all relevant examples.  Also permits use of "1" as an alternative toggle key on certain platforms if function keys not available or a screen keyboard is used (iOS/macOS, Android).
2. Implements ImGui UI overlay for dynamicrendering example.  Previously this example crashed on linux and macOS when not in benchmarking mode (i.e. when UI overlay was set to visible but not implemented).  This fixes the crash and implements the UI overlay for this example on all platforms. Also adds `VK_KHR_dynamic_rendering` extension dependencies to eliminate validation errors.
3. Moves macOS benchmarking to the NSApp rendering loop, which makes the example window visible (previously was headless when in benchmarking mode on macOS).  This aligns macOS benchmarking behaviour with Windows and Linux.

Tested on macOS, iOS, Windows, Linux. Improvements for Android (F1/1/L1 button UI overlay toggle and support for textoverlay spacebar UI toggle) were implemented but could not be tested since I do not have that platform.  Would be great if @SaschaWillems could verify these improvements on Android.

### DIFF
--- a/base/VulkanUIOverlay.cpp
+++ b/base/VulkanUIOverlay.cpp
@@ -217,7 +217,7 @@ namespace vks
 	}
 
 	/** Prepare a separate pipeline for the UI overlay rendering decoupled from the main application */
-	void UIOverlay::preparePipeline(const VkPipelineCache pipelineCache, const VkRenderPass renderPass)
+	void UIOverlay::preparePipeline(const VkPipelineCache pipelineCache, const VkRenderPass renderPass, const VkFormat colorFormat, const VkFormat depthFormat)
 	{
 		// Pipeline layout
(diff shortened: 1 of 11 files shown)


SaschaWillems:
Thanks again for your PR. Ver much appreciated.

The changes look good to me, and work as expected on Android.

SRSaunders:
Thanks again for merging and appreciate your testing on Android.

PR #991, opened 2022-09-08
Example demonstrating how to use SBT record data in a ray tracing pipeline

natevm:
I noticed that none of the ray tracing examples in your repo shows how to use SBT record data. This PR adds a little example showing how you can upload data between shader handles that can be accessed on the device.


SaschaWillems:
Thank you very much for your PR. This is very much appreciated!

It'll probably take a week or two for me to review this though.

natevm:
Hi Sascha, 
Let me know if there's anything you'd like me to add to the example. 

I'm thinking I might also submit a PR also demonstrating how to use custom intersection shaders and AABB trees...

examples/raytracingsbtdata/raytracingsbtdata.cpp
```diff
@@ -0,0 +1,922 @@
+/*
+* Vulkan Example - Hardware accelerated ray tracing example using SBT data
+*
+* Uses the data section of each shader binding table record to color the background and geometry
+*
+* Copyright (C) 2019-2020 by Sascha Willems - www.saschawillems.de
```
SaschaWillems, line 6: Feel free to add your name here as a contributor.

natevm:
Ok, added my name for reference to the example. I also added some comments that will help explain what's going on with the data section of the SBT records, and I modified the setup slightly to incorporate the miss record's data as well.

SaschaWillems:
The sample runs fine for me, but I get a few validation layer errors:

```ERROR: [-836035013][VUID-vkCmdTraceRaysKHR-size-04023] : Validation Error: [ VUID-vkCmdTraceRaysKHR-size-04023 ] Object 0: handle = 0x1af66b20ff0, type = VK_OBJECT_TYPE_DEVICE; | MessageID = 0xce2b1e3b | vkCmdTraceRaysKHR: The size member of pRayGenShaderBindingTable must be equal to its stride member The Vulkan spec states: The size member of pRayGenShaderBindingTable must be equal to its stride member (https://www.khronos.org/registry/vulkan/specs/1.3-extensions/html/vkspec.html#VUID-vkCmdTraceRaysKHR-size-04023)
```

Setting the size to the same value as the stride gets rid of the message and the sample still works, but I'm not sure if that's intended. 

Can you take a look at this and also maybe add a screenshot of how the sample is supposed to look in here?

PR #993, opened 2022-09-14
Fixes to dynamic state behavior in pipelines demo

MTCoster:
This patch includes three fixes:
 - Allow `wideLines` to be enabled when `fillModeNotSolid` isn't,
 - Check against `enabledFeatures` for runtime conditionals instead of the theoretical values in `deviceFeatures`, and
 - Always call `vkCmdSetLineWidth()` before any draw command occurs. This is required by the spec, since any state declared as dynamic is explicitly not initialized by `VkPipeline*StateCreateInfo`.

```
Signed-off-by: Matt Coster <matt.coster@imgtec.com>
```


SaschaWillems:
Thank you very much for this PR. Very much appreciated!

PR #994, opened 2022-09-15
Update viewportarray.cpp

tanliping2011:
GPU does not support multi viewports and exit directly.


SaschaWillems:
`vks::tools::exitFatal` already includes an exit with the exact same error code, so I don't think this is necessary at all.

PR #999, opened 2022-11-21
dynamicrendering: use the right layout for depth/stencil

djdeath:


SaschaWillems:

Issue #665, opened 2020-02-07
Misleading example base subpass dependencies?


krOoze:
```

	dependencies[0].srcSubpass = VK_SUBPASS_EXTERNAL;
	dependencies[0].dstSubpass = 0;
	dependencies[0].srcStageMask = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;
	dependencies[0].dstStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
	dependencies[0].srcAccessMask = VK_ACCESS_MEMORY_READ_BIT;
	dependencies[0].dstAccessMask = VK_ACCESS_COLOR_ATTACHMENT_READ_BIT | VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
	dependencies[0].dependencyFlags = VK_DEPENDENCY_BY_REGION_BIT;
```

Not entirelly sure why exacly it is `VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT` (blocking _execution dependency_), so it probably needs a comment. Probably it should be either `ALL_COMMANDS` and have `WRITE` too, or it should be `TOP_OF_PIPE`  and the access mask should be `0`.

Furthemore, the `loadOp` is set as `CLEAR`, so only `VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT` should be needed.

```
	dependencies[1].srcSubpass = 0;
	dependencies[1].dstSubpass = VK_SUBPASS_EXTERNAL;
	dependencies[1].srcStageMask = VK_PIPELINE_STAGE_COLOR_ATTACHMENT_OUTPUT_BIT;
	dependencies[1].dstStageMask = VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT;
	dependencies[1].srcAccessMask = VK_ACCESS_COLOR_ATTACHMENT_READ_BIT | VK_ACCESS_COLOR_ATTACHMENT_WRITE_BIT;
	dependencies[1].dstAccessMask = VK_ACCESS_MEMORY_READ_BIT;
	dependencies[1].dependencyFlags = VK_DEPENDENCY_BY_REGION_BIT;
```

`VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT` has no memory access. The wildcard `VK_ACCESS_MEMORY_READ_BIT` is technically allowed, though the access mask should probably be `0`

[...]

PR #1006, opened 2023-01-05
Fix for base subpass depedencies had uninitialized dependency flags.

per-mathisen-arm:
This consistently crashed all the examples on my linux laptop with Intel GPU in it.

Ref #665


SaschaWillems:
Thanks a lot for this PR. I embarrassingly forgot to zero initialize the dependencies.

PR #1014, opened 2023-02-19
Use the secure https URL for assets

skal65535:
Otherwise you get 'access denied'.

### DIFF
--- a/download_assets.py
+++ b/download_assets.py
@@ -4,7 +4,7 @@ import sys
 from urllib.request import urlretrieve
 from zipfile import ZipFile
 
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Can you elaborate? This works fine for me, and I haven't heard of any problems before this. What http error code do you get? And are you behind some corporate firewall or proxy?

skal65535:
hmm... I *think* i was getting a 403 error. And it wasn't on my corporate network. But it seems to not be reproducible any more. Strange...

Well, up to you if want to switch to https or not.
Thanks!

PR #1025, opened 2023-04-11
fix getOffset for sampler

RYDB3RG:


SaschaWillems:
Thank you very much for your PR. But since I'm not using the offsets at all in my sample, I instead just removed those lines.

PR #1026, opened 2023-04-12
Add in missing screenshots for examples

tek256:
Fixes a few existing READMEs, adds a bunch more, and adds missing screenshots for others


SaschaWillems:
A quick note on PRs like this: It's always a good idea to ask up-front if such a PR makes sense. It contains screenshots from samples that I don't consider finished yet, and tbh I stopped adding screenshots for a reason, and may remove them completely at some point due to the fact that samples may change over time.

PR #1028, opened 2023-04-18
Use EXT for mesh shader feature

K1ngst0m:
Replace `_NV` to `_EXT` for supporting other platforms


SaschaWillems:
Thanks. Your PR is very much appreciated.

PR #1038, opened 2023-05-29
Building for Android Improvement

robotchaoX:
Hi owner,
Here are 4 minor changes for Andorid building. Would you accept it?

[Increase the JVM heap size for Gradle]
[Why]
The Java Virtual Machine(JVM) maximum heap size(-Xmx) is 512M by default. It will occur out of memory eror when compiling for Android.
Android compiling error: "java.lang.OutOfMemoryError: Java heap space".
[How]
Configure JVM memory for Gradle by setting Gradle property org.gradle.jvmargs in gradle.properties file.
This gradle.properties file was generated by Android Studio settings.
Reference:
https://developer.android.com/build/optimize-your-build?utm_source=android-studio#increase-the-jvm-heap-size
https://docs.gradle.org/7.2/userguide/build_environment.html#sec:configuring_jvm_memory

[Migrate jcenter to mavenCentral]
[Why]
JFrog made JCenter a read-only repository on March 31, 2021.
[How]
Migrate JCenter to Maven Central.
Reference: https://developer.android.com/build/jcenter-migration

[Update BUILD.md for building on Android]

[set abiFilters via a global variable in the settings.gradle]

Todo
~~[Add x86_64 ABI support for Android]
[Why]
In some case, people may need x86 ABI version apk if they develop on x86 Android.
Support x86_64 ABI by default is better.
[How]
Add android::ndk::abiFilters property "x86_64".~~


SaschaWillems:
Is it possible to turn the x86_64 abi into an option? x86_64 has a very small market share, but always having to also built this abi will increase built time. I'd prefer to have this set via a global variable in the settings.gradle file (see e.g. https://github.com/SaschaWillems/Vulkan/blob/240a56e44108d480bb3b01444b26552ba70857cc/android/build.gradle#L28) so people that want to compile for that abi simply add it there.

robotchaoX:
Hi SaschaWillems,
I set abiFilters via global variable in settings.gradle as you required. Will find the way to add x86_64 ABI support through project property. It may take time. Did other commits meet your requirements?

robotchaoX:
Hi SaschaWillems,
Change the topic to "Building for Android Improvement". Do current commits meet your requirements?
If I find the way to add x86_64 ABI support through project property will create another PR.

SaschaWillems:
Yes, that looks a lot better and makes it very easy to adjust abis. Will try to review it this weekend.

BUILD.md
```diff
@@ -23,7 +23,11 @@ Use the provided CMakeLists.txt with [CMake](https://cmake.org) to generate a bu
 
 Building on Android is done using the [Gradle Build Tool](https://gradle.org/):
 
+Set Android SDK/NDK by environment variable `ANDROID_SDK_ROOT`/`ANDROID_NDK_HOME`.
+
 ```
+export ANDROID_SDK_ROOT=<path-to-SDK>
```
SaschaWillems, line 29: This only applies to Linux I guess. Can you take this oustide of the ode block and add a note that this is only required on that platform?

BUILD.md
```diff
@@ -33,6 +37,8 @@ On Windows execute `gradlew.bat assembleDebug`.
 
 If you want to build and install on a connected device or emulator image, run ```gradle installDebug``` instead.
 
+If you want to build it through [Android Studio](https://developer.android.com/studio), open project ```android\build.gradle``` in Android Studio.
```
SaschaWillems, line 40: That's not 100% correct. At least on Windows you just select the android folder:

image

SaschaWillems (changes requested):

SaschaWillems (changes requested):
Looks mostly good to me. A added two comments regarding the changes to the build.md.

BUILD.md
robotchaoX: Change to just select the android folder.

BUILD.md
robotchaoX: Thanks for your reminding. The user should know the way how to set environment variables on different platform. Delete that export command.

robotchaoX (commented):
I modified it according to your two suggestions.

PR #1049, opened 2023-07-10
Fix install location for assets and shaders

thochstein:
This fixes the following error for the cross-compile case, where RESOURCE_INSTALL_DIR=/usr/share/vulkan-demos:
```
Error: Could not open shader file \"/usr/share/vulkan-demos/shaders/glsl/renderheadless/triangle.frag.spv\"
```


SaschaWillems:

PR #1058, opened 2023-08-22
Fix a comment in shadowmappingcascade example

ZaOniRinku:
When making the shadowmaps image and image views, a comment said "One image and framebuffer per cascade", while it should be "One image **view** and framebuffer per cascade". The code is right.


ZaOniRinku:
Also kind of unrelated I guess? In the "updateCascades()" function, aren't these calculations redundant?
https://github.com/SaschaWillems/Vulkan/blob/master/examples/shadowmappingcascade/shadowmappingcascade.cpp#L640-L647

clipRange = farClip - nearClip
minZ = nearClip
maxZ = nearClip + clipRange = nearClip + (farClip - nearClip) = farClip
range = maxZ - minZ = farClip - nearClip = clipRange

PR #1062, opened 2023-09-02
Clarify comment to specify where exactly we do this command buffer "creation".

Jerrody:


SaschaWillems:
Tbh I'm happy with the current wording.

PR #1074, opened 2023-09-21
Specify android:exported explicitly for <activity>

robotchaoX:
[Why]
Make it compatible with targetSdkVersion >= 31.
android:exported needs to be explicitly specified for <activity>. Apps targeting Android 12 and higher are required to specify an explicit value for `android:exported` when the corresponding component has an intent filter defined. See https://developer.android.com/guide/topics/manifest/activity-element#exported for details.

[How]
Specify android:exported explicitly for <activity>.


robotchaoX:
Hi owner,
If you set targetSdkVersion >= 31, there will be compiling error for Android. Because there are more restriction for Android 12.
This commit make it compatible with targetSdkVersion >= 31.

SaschaWillems:
A small request: Please ask before doing such large PRs. This clashes with something I've bee doing locally, so I probably won't be able to merge this.

PR #1075, opened 2023-10-03
Fixed 2 separate issues. 

scottchou007:
Please refer to commit descriptions for details.

### DIFF
--- a/base/VulkanTools.cpp
+++ b/base/VulkanTools.cpp
@@ -430,7 +430,7 @@ namespace vks
 			std::ifstream f(filename.c_str());
 			return !f.fail();
 		}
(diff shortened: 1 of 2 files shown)


base/VulkanTools.cpp
```diff
@@ -430,16 +430,15 @@ namespace vks
 			std::ifstream f(filename.c_str());
 			return !f.fail();
 		}
-
+#if !defined(_WIN32)
```
SaschaWillems, line 433: Can you elaborate why this is required? Windows is my main platform and I never had any troubles with this.

SaschaWillems (changes requested):

PR #1076, opened 2023-10-04
Fix 2 separate issues. 

scottchou007:
1st commit:
This project's default build config is x64. When building with 32-bit Windows config, there will be the following errors:
"Vulkan\base\VulkanTools.cpp(440): error C2084: function 'uint32_t vks::tools::alignedSize(uint32_t,uint32_t)' already has a body"
This change fixed it, however if there is no plan to support 32-bit Windows build, this change is not necessary.

2nd commit:
Fixed some render problems on certain platforms.

### DIFF
--- a/base/VulkanTools.cpp
+++ b/base/VulkanTools.cpp
@@ -430,16 +430,15 @@ namespace vks
 			std::ifstream f(filename.c_str());
 			return !f.fail();
 		}
-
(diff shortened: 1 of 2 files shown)


scottchou007:
Created new pull to replace the old one, content revised and commits squashed.

PR #1077, opened 2023-10-05
fixed 'descriptor set layout' bug for textures in 'gltfloading' example

nelson-steve:
In "gltfloading" example 'descriptor set layout' for textures was being created with the same 'descriptor set layout create info' as matrices(UBO) which is wrong.
It seems that line of code was missing.


SaschaWillems:
Please provide a proper description.

nelson-steve:
> Please provide a proper description.

I've done it. Thanks

SaschaWillems:
That fix isn't necessary.

This:

```cpp
VkDescriptorSetLayoutCreateInfo descriptorSetLayoutCI = vks::initializers::descriptorSetLayoutCreateInfo(&setLayoutBinding, 1);
```` 
Stores a pointer to `setLayoutBinding`.

So the change of `setLayoutBinding` in line 619 will update the reference in the descriptor set layout create info.

nelson-steve:
Yeah. You're right. :)

PR #1082, opened 2023-10-31
Fix parallaxmapping asset missing on Android

robotchaoX:
[Why]
Asset name plane.gltf is not corrected in gradle copyTask on Android.

[How]
Correct the sample parallaxmapping's asset name plane.gltf in build.gradle.


robotchaoX:
Hi owner,
Please verify this tiny fix. Thanks!

PR #1084, opened 2023-11-01
Generate example variablerateshading for Android

robotchaoX:
[Why]
Example variablerateshading was not compiled for Android.

[How]
Generate Android version example variablerateshading from android/examples/_template.


SaschaWillems:

PR #1086, opened 2023-11-15
Warning Fixes

LukasBuricin:
First of all, let me thank you for the great amount of effort whilst making these examples.

This Pull Request addresses most of the warnings when building in x64 configuration.

The warning types addressed:
- Implicit conversion from `size_t` to `uint32_t`
- Implicit conversion to `float`
- Comparison of singed with unsigned
- Unused local variables

Also changed `NULL` to `nullptr` when changing the line.

I hope you find this useful.

### DIFF
--- a/examples/descriptorbuffer/descriptorbuffer.cpp
+++ b/examples/descriptorbuffer/descriptorbuffer.cpp
@@ -204,8 +204,8 @@ public:
 		vkGetDescriptorSetLayoutBindingOffsetEXT(device, combinedImageDescriptor.setLayout, 0, &combinedImageDescriptor.layoutOffset);
 
 		// In order to copy resource descriptors to the correct place, we need to calculate aligned sizes
-		uniformDescriptor.layoutSize = vks::tools::alignedSize(uniformDescriptor.layoutSize, descriptorBufferProperties.descriptorBufferOffsetAlignment);
-		combinedImageDescriptor.layoutSize = vks::tools::alignedSize(combinedImageDescriptor.layoutSize, descriptorBufferProperties.descriptorBufferOffsetAlignment);
+		uniformDescriptor.layoutSize = vks::tools::alignedSize(static_cast<uint32_t>(uniformDescriptor.layoutSize), static_cast<uint32_t>(descriptorBufferProperties.descriptorBufferOffsetAlignment));
+		combinedImageDescriptor.layoutSize = vks::tools::alignedSize(static_cast<uint32_t>(combinedImageDescriptor.layoutSize), static_cast<uint32_t>(descriptorBufferProperties.descriptorBufferOffsetAlignment));
 
 		// This buffer will contain resource descriptors for all the uniform buffers (one per cube and one with global matrices)
 		VK_CHECK_RESULT(vulkanDevice->createBuffer(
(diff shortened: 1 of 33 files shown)


examples/descriptorbuffer/descriptorbuffer.cpp
```diff
@@ -204,8 +204,8 @@ class VulkanExample : public VulkanExampleBase
 		vkGetDescriptorSetLayoutBindingOffsetEXT(device, combinedImageDescriptor.setLayout, 0, &combinedImageDescriptor.layoutOffset);
 
 		// In order to copy resource descriptors to the correct place, we need to calculate aligned sizes
-		uniformDescriptor.layoutSize = vks::tools::alignedSize(uniformDescriptor.layoutSize, descriptorBufferProperties.descriptorBufferOffsetAlignment);
-		combinedImageDescriptor.layoutSize = vks::tools::alignedSize(combinedImageDescriptor.layoutSize, descriptorBufferProperties.descriptorBufferOffsetAlignment);
+		uniformDescriptor.layoutSize = vks::tools::alignedSize(static_cast<uint32_t>(uniformDescriptor.layoutSize), static_cast<uint32_t>(descriptorBufferProperties.descriptorBufferOffsetAlignment));
+		combinedImageDescriptor.layoutSize = vks::tools::alignedSize(static_cast<uint32_t>(combinedImageDescriptor.layoutSize), static_cast<uint32_t>(descriptorBufferProperties.descriptorBufferOffsetAlignment));
```
LukasBuricin, line 208: For consideration: Maybe an additional prototype

    VkDeviceSize vks::tools::alignedSize(VkDeviceSize, VkDeviceSize);

would be useful, since all the involved types here are `VkDeviceSize`. In that case the casting would not be necessary in this case.

examples/gears/vulkangear.cpp
```diff
@@ -53,13 +53,13 @@ void VulkanGear::generate(GearInfo *gearinfo, VkQueue queue)
 	r0 = gearinfo->innerRadius;
 	r1 = gearinfo->outerRadius - gearinfo->toothDepth / 2.0f;
 	r2 = gearinfo->outerRadius + gearinfo->toothDepth / 2.0f;
-	da = 2.0f * M_PI / gearinfo->numTeeth / 4.0f;
+	da = static_cast<float>(2.0f * M_PI / gearinfo->numTeeth / 4.0f);
```
LukasBuricin, line 56: For consideration: When using C++20 this and other places could be using `std::numbers::pi_v<float>` instead of the `M_PI` which is a `#define` and considered somewhat obsolete. Also, with `std::numbers::pi_v<float>` the cast would not be needed.

LukasBuricin (commented):
There is one more place I didn't touch in

    void VulkanExample::generateTextures()

causing warnings about truncating `short` to `uint8_t`.

My suggesting would be changing:

    std::uniform_int_distribution<short> rndDist(50, 255);

to

    std::uniform_int_distribution<uint8_t> rndDist(50, 255);

I just wanted to clarify for case I am missing something, but if you agree I can make the change.

Thank you again!

SaschaWillems:
Thank you very much for your PR. As this is a rather large PR please give me some time to review it.

examples/gears/vulkangear.cpp
SaschaWillems: I don't plan on using C++ 20. The samples support a wide range of platforms incl. some not widely used outside of some niches. And some of those lack behind with newer C++ features.

PR #1095, opened 2023-12-31
Fixed image link

dustContributor:
It seems the image doesn't has underscores in its name.


SaschaWillems:

PR #1098, opened 2024-01-11
Use debug utils during instance creation

jherico:
The `VulkanDebug.h` header is referencing the old `VK_EXT_debug_report` message callback signature, which no longer exists.  I've updated the declaration to match the  `VK_EXT_debug_utils` signature and changed the name in `VulkanDebug.cpp` to match.

Additionally, if validation is enabled, it should be applied to instance creation as well as everything after that.  This is allowed by `VK_EXT_debug_utils` by passing the `VkDebugUtilsMessengerCreateInfoEXT` structure in the `pNext` chain of the `VkInstanceCreateInfo` structure, which the base example code now does.

### DIFF
--- a/base/VulkanDebug.cpp
+++ b/base/VulkanDebug.cpp
@@ -17,7 +17,7 @@ namespace vks
 		PFN_vkDestroyDebugUtilsMessengerEXT vkDestroyDebugUtilsMessengerEXT;
 		VkDebugUtilsMessengerEXT debugUtilsMessenger;
 
-		VKAPI_ATTR VkBool32 VKAPI_CALL debugUtilsMessengerCallback(
+		VKAPI_ATTR VkBool32 VKAPI_CALL debugUtilsMessageCallback(
 			VkDebugUtilsMessageSeverityFlagBitsEXT messageSeverity,
 			VkDebugUtilsMessageTypeFlagsEXT messageType,
 			const VkDebugUtilsMessengerCallbackDataEXT* pCallbackData,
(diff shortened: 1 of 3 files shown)


SaschaWillems:
Thank you very much for your contribution.

base/VulkanDebug.cpp
```diff
@@ -17,7 +17,7 @@ namespace vks
 		PFN_vkDestroyDebugUtilsMessengerEXT vkDestroyDebugUtilsMessengerEXT;
 		VkDebugUtilsMessengerEXT debugUtilsMessenger;
 
-		VKAPI_ATTR VkBool32 VKAPI_CALL debugUtilsMessengerCallback(
```
SaschaWillems, line 20: Thanks for spotting this!

base/vulkanexamplebase.cpp
```diff
@@ -98,6 +98,22 @@ VkResult VulkanExampleBase::createInstance(bool enableValidation)
 	instanceCreateInfo.pNext = NULL;
 	instanceCreateInfo.pApplicationInfo = &appInfo;
 
+
+
+	VkDebugUtilsMessengerCreateInfoEXT debugUtilsMessengerCI{};
```
SaschaWillems, line 103: Would it be possible to somehow merge this with the code from `setupDebugging` in `VulkanDebug.h`, which already sets up the debug messenger? Otherwise this results in code duplication. I'd also remove some verbosity flags and message types like in `setupDebugging`.

SaschaWillems (changes requested):

base/vulkanexamplebase.cpp
jherico: Technically I could create a `vks::debug::setupInstanceCreateDebugging(VkInstanceCreateInfo& instanceCreateInfo)` function, but the `VkDebugUtilsMessengerCreateInfoEXT` that goes into the `instanceCreateInfo.pNext`  value needs to live beyond the scope of such a function.  it has to exist until vkCreateIntance is called and returns.  I suppose I could declare it as static isnide the `VulkanDebug.cpp` unit, but that seems weird to me.

I've tried to strike a balance with my new commit where the `VkDebugUtilsMessengerCreateInfoEXT` is still declared in `vulkanexamplebase`, but all the boilerplate for populating it is moved to a function in `VulkanDebug.cpp` that is also then used by the existing code to setup debugging.

Hope this is satisfactory, but I can switch to a static declaration if you'd prefer.

base/VulkanDebug.cpp
jherico: Weird. I honestly didn't mean to make that change.  I think I changed both to `messageCallback` and then thought better of it and tried to rename it back but started with the header and misremembered the name.

SaschaWillems:
This looks good time. Thank you very much 👍🏻

PR #1099, opened 2024-01-16
fix a small bug in pbr-ibl

F-Mu:
Because of the length of vec4 is also 4,it runs as expected.


SaschaWillems:

PR #1100, opened 2024-01-25
Remove some unused variables, less verbose animation logic

jherico:
This adds a `Random` utility class that provides more functionality than the existing `rnd()` function used in the multithreaded example.  

It does some minor cleanup on the animation logic.  

* `glm::vec3 rotation` replaced by `float rotationAxis;` because the X and Z values are never read or written.
* `glm::mat4 model` removed as it's only used as an intermediate value to write to the push constant structure
* The push constant structure is absorbed by ObjectData because there's no reason for separate storage, they're always 1:1
* Logic to initialize `ObjectData`, update it per-frame and extract the mat4 are moved into the structure itself, reducing the verbosity and amount of indirection required to access all the variables
* Where possible convert values to radians at initialization time rather than each update
* In a couple places where the logic was something like `x = a + rnd(b)` I've updated it to be `random.real(a, a+b)` which provides the same range of values but feels easier to read
* Eliminated unused variables stateT, posX and posZ

### DIFF
--- a/examples/multithreading/multithreading.cpp
+++ b/examples/multithreading/multithreading.cpp
@@ -126,7 +126,6 @@ public:
 
 	VulkanExample() : VulkanExampleBase()
 	{
-		// settings.validation = true;
 		title = "Multi threaded command buffer";
 		camera.type = Camera::CameraType::lookat;
 		camera.setPosition(glm::vec3(0.0f, -0.0f, -32.5f));
@@ -135,6 +134,7 @@ public:
 		camera.setPerspective(60.0f, (float)width / (float)height, 0.1f, 256.0f);
 		// Get number of max. concurrent threads
 		numThreads = std::thread::hardware_concurrency();
+		assert(numThreads > 0);
 #if defined(__ANDROID__)
 		LOGD("numThreads = %d", numThreads);
 #else
--- a/examples/multithreading/multithreading.cpp
+++ b/examples/multithreading/multithreading.cpp
@@ -10,8 +10,11 @@
 
 #include "threadpool.hpp"
 #include "frustum.hpp"
-
+#include "random.hpp"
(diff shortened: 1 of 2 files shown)


examples/multithreading/multithreading.cpp
```diff
@@ -58,23 +61,53 @@ class VulkanExample : public VulkanExampleBase
...
+			rotationAxis += 2.5f * rotationSpeed * frameTimer;
+			rotationAxis = fmod(rotationAxis, M_TAU);
+			deltaT += 0.15f * frameTimer;
+			deltaT = fmod(deltaT, 1.0f);
+			pos.y = sin(deltaT * M_TAU) * 2.5f;
+		}
+
+		glm::mat4 getModel() {
+			auto model = glm::translate(glm::mat4(1.0f), pos);
+			model = glm::rotate(model, -sinf(deltaT * M_TAU) * 0.25f, glm::vec3(rotationDir, 0.0f, 0.0f));
```
jherico, line 98: I've reproduced the original logic here, but I'm confused.  `glm::rotate` expects an angle here, in radiants, but we're passing the output of a `sin` function.  Two lines below we pass in `deltaT * M_TAU` without wrapping it in a `sin`.  I'm not sure what the intent is.

SaschaWillems:
I'll probably remote that sample all along. It was never a good sample after all.

jherico:
I mean, I like the multithreaded sample, but OK.

SaschaWillems:
IMO it was kinda okay back when I launched the samples, but that changed with CPUs getting faster. On many setups it's not slower than doing the same in a single thread, and I've been pondering for years on doing a proper up-to-date and more real-world like multi threading sample. But we do have one over at the Khronos sample anyway.

PR #1103, opened 2024-02-05
Fix barrier issues in computecloth

### LINKED ISSUE #1097: Unnecessary barrier + possibly incorrect barriers?
You're moving both the storage buffers between the compute and graphics queues and vice versa, but only the `storageBuffers.output.buffer` is actually used on the graphics queue.

https://github.com/SaschaWillems/Vulkan/blob/44ff7a1a9dc51e8d70f5ee9e4d6175fdc2c36d82/examples/computecloth/computecloth.cpp#L188

Looking more closely at this, I think the logic in `addComputeToComputeBarriers` is wrong.  The loop in `buildComputeCommandBuffer` executes 64 times, using the descriptor set to swap which buffer is being read and which one is being written, but the `addComputeToComputeBarriers` only ever does one thing... it ensures that both the input and output buffers are going from a write state to a read state.

One barrier should be going from read to write, and the other should be going from write to read, and the next frame they should swap back, over and over until the loop is exited, at which point the cross-queue barriers come into play.

jherico:
This is to fix #1097

Notes inline

The updated sample passes with the Vulkan Configurator set to enable the validation profile and with it set to the synchronization profile, with no errors or warnings.

### DIFF
--- a/base/VulkanInitializers.hpp
+++ b/base/VulkanInitializers.hpp
@@ -90,7 +90,7 @@ namespace vks
 			return imageMemoryBarrier;
 		}
 
-		/** @brief Initialize a buffer memory barrier with no image transfer ownership */
+		/** @brief Initialize a buffer memory barrier with no buffer transfer ownership */
 		inline VkBufferMemoryBarrier bufferMemoryBarrier()
 		{
 			VkBufferMemoryBarrier bufferMemoryBarrier {};
--- a/examples/computecloth/computecloth.cpp
+++ b/examples/computecloth/computecloth.cpp
@@ -16,7 +16,6 @@
 class VulkanExample : public VulkanExampleBase
 {
 public:
-	uint32_t readSet{ 0 };
 	uint32_t indexCount{ 0 };
 	bool simulateWind{ false };
 	// This will be set to true, if the device has a dedicated queue from a compute only queue family
@@ -39,6 +38,7 @@ public:
 	struct Cloth {
 		glm::uvec2 gridsize{ 60, 60 };
 		glm::vec2 size{ 5.0f, 5.0f };
+		uint32_t numParticles{ gridsize.x * gridsize.y };
 	} cloth;
 
 	// We put the resource "types" into structs to make this sample easier to understand
@@ -77,7 +77,7 @@ public:
 		} semaphores;
 		VkQueue queue{ VK_NULL_HANDLE };
 		VkCommandPool commandPool{ VK_NULL_HANDLE };
-		std::array<VkCommandBuffer, 2> commandBuffers{};
+		std::array<VkCommandBuffer, 1> commandBuffers{ VK_NULL_HANDLE };
 		VkDescriptorSetLayout descriptorSetLayout{ VK_NULL_HANDLE };
 		std::array<VkDescriptorSet, 2> descriptorSets{ VK_NULL_HANDLE };
 		VkPipelineLayout pipelineLayout{ VK_NULL_HANDLE };
@@ -160,35 +160,34 @@ public:
 			bufferBarrier.srcQueueFamilyIndex = vulkanDevice->queueFamilyIndices.graphics;
 			bufferBarrier.dstQueueFamilyIndex = vulkanDevice->queueFamilyIndices.compute;
 			bufferBarrier.size = VK_WHOLE_SIZE;
-
-			std::vector<VkBufferMemoryBarrier> bufferBarriers;
-			bufferBarrier.buffer = storageBuffers.input.buffer;
-			bufferBarriers.push_back(bufferBarrier);
 			bufferBarrier.buffer = storageBuffers.output.buffer;
-			bufferBarriers.push_back(bufferBarrier);
 			vkCmdPipelineBarrier(commandBuffer,
 				srcStageMask,
 				dstStageMask,
 				VK_FLAGS_NONE,
 				0, nullptr,
-				static_cast<uint32_t>(bufferBarriers.size()), bufferBarriers.data(),
+				1, &bufferBarrier,
 				0, nullptr);
 		}
 	}
 
-	void addComputeToComputeBarriers(VkCommandBuffer commandBuffer)
+	void addComputeToComputeBarriers(VkCommandBuffer commandBuffer, bool reverse)
 	{
-		VkBufferMemoryBarrier bufferBarrier = vks::initializers::bufferMemoryBarrier();
-		bufferBarrier.srcAccessMask = VK_ACCESS_SHADER_WRITE_BIT;
-		bufferBarrier.dstAccessMask = VK_ACCESS_SHADER_READ_BIT;
-		bufferBarrier.srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
-		bufferBarrier.dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
-		bufferBarrier.size = VK_WHOLE_SIZE;
-		std::vector<VkBufferMemoryBarrier> bufferBarriers;
-		bufferBarrier.buffer = storageBuffers.input.buffer;
-		bufferBarriers.push_back(bufferBarrier);
-		bufferBarrier.buffer = storageBuffers.output.buffer;
-		bufferBarriers.push_back(bufferBarrier);
+		std::array<VkBufferMemoryBarrier, 2> bufferBarriers;
+		bufferBarriers[0] = vks::initializers::bufferMemoryBarrier();
+		bufferBarriers[0].srcAccessMask = VK_ACCESS_SHADER_READ_BIT;
+		bufferBarriers[0].dstAccessMask = VK_ACCESS_SHADER_WRITE_BIT;
+		bufferBarriers[0].buffer = storageBuffers.input.buffer;
+		bufferBarriers[0].size = VK_WHOLE_SIZE;
+		bufferBarriers[1] = vks::initializers::bufferMemoryBarrier();
+		bufferBarriers[1].srcAccessMask = VK_ACCESS_SHADER_WRITE_BIT;
+		bufferBarriers[1].dstAccessMask = VK_ACCESS_SHADER_READ_BIT;
+		bufferBarriers[1].buffer = storageBuffers.output.buffer;
+		bufferBarriers[1].size = VK_WHOLE_SIZE;
+		if (reverse) {
+			std::swap(bufferBarriers[0].buffer, bufferBarriers[1].buffer);
+		}
+
 		vkCmdPipelineBarrier(
 			commandBuffer,
 			VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT,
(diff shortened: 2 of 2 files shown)


base/VulkanDevice.cpp
```diff
@@ -490,6 +490,9 @@ namespace vks
 		if (begin)
 		{
 			VkCommandBufferBeginInfo cmdBufInfo = vks::initializers::commandBufferBeginInfo();
+			// If we're allocating AND starting the command buffer, we're almost certainly going to 
+			// use it right away, so we'll use the ONE_TIME_SUBMIT_BIT flag
+			cmdBufInfo.flags |= VK_COMMAND_BUFFER_USAGE_ONE_TIME_SUBMIT_BIT;
```
jherico, line 495: I see no uses of  `vks::VulkanDevice::createCommandBuffer(..., true)` that aren't followed by some one off code and a vks::VulkanDevice::flushCommandBuffer` with the `free` parameter implicitly set tot true (I also see no uses of `vks::VulkanDevice::flushCommandBuffer` with flush set to false.  Have you considered a

```c++
vks::VulkanDevice::withOneTimeCommandBuffer(
    VkQueue, 
    VkCommandPool, 
    const std::function<void(VkCommandBuffer)>&);
```
function to encapsulate those `createCommandBuffer` / `flushCommandBuffer` pairs?

base/VulkanInitializers.hpp
```diff
@@ -90,13 +90,14 @@ namespace vks
...
 
-		/** @brief Initialize a buffer memory barrier with no image transfer ownership */
+		/** @brief Initialize a buffer memory barrier with no buffer transfer ownership */
 		inline VkBufferMemoryBarrier bufferMemoryBarrier()
 		{
 			VkBufferMemoryBarrier bufferMemoryBarrier {};
 			bufferMemoryBarrier.sType = VK_STRUCTURE_TYPE_BUFFER_MEMORY_BARRIER;
 			bufferMemoryBarrier.srcQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
 			bufferMemoryBarrier.dstQueueFamilyIndex = VK_QUEUE_FAMILY_IGNORED;
+			bufferMemoryBarrier.size = VK_WHOLE_SIZE;
```
jherico, line 100: `VkBufferMemoryBarrier::size` being zero is results in an invalid barrier, so everywhere you're using this helper, you must be assigning to the size, meaning this will always be overwritten if you're being explicit about the size (and is therefore innocuous as a change) but it means everywhere you're explicitly setting it to `VK_WHOLE_SIZE` could eliminate that line.

examples/computecloth/computecloth.cpp
```diff
@@ -77,7 +76,7 @@ class VulkanExample : public VulkanExampleBase
 		} semaphores;
 		VkQueue queue{ VK_NULL_HANDLE };
 		VkCommandPool commandPool{ VK_NULL_HANDLE };
-		std::array<VkCommandBuffer, 2> commandBuffers{};
+		VkCommandBuffer commandBuffer{ VK_NULL_HANDLE };
```
jherico, line 79: I _think_ I get why this was two command buffers.  You were trying to set it up so that at the end of each compute loop, you took whichever SSBO had been written last and used it as the input to the next cycle?  But even though you transition both SSBOs to the graphics queue, the input buffer is _never_ used as a vertex buffer in the original code, and there aren't two corresponding graphics command buffers / descriptor sets to switch between using storageBuffer.input and storageBuffer.output as an input like there are in compute.

base/VulkanInitializers.hpp
SaschaWillems: This is only the initializer. All places that use it set the buffer size afterwards.

base/VulkanDevice.cpp
SaschaWillems: Aside from trying to fix sync, my samples are in maintenance mode. I don't plan on putting a lot of work into them in the future. I mostly work on the Khronos samples.

examples/computecloth/computecloth.cpp
```diff
@@ -292,8 +287,7 @@ class VulkanExample : public VulkanExampleBase
 		VkCommandBufferBeginInfo cmdBufInfo = vks::initializers::commandBufferBeginInfo();
 		cmdBufInfo.flags = VK_COMMAND_BUFFER_USAGE_SIMULTANEOUS_USE_BIT;
 
-		for (uint32_t i = 0; i < 2; i++) {
-
+		for (uint32_t i = 0; i < compute.commandBuffers.size(); i++) {
```
jherico, line 290: This feels kind of dumb since it's only a single item array now, but Github failed it's saving throw in producing a reasonable diff when I removed the for loop.

examples/computecloth/computecloth.cpp
```diff
@@ -306,26 +300,66 @@ class VulkanExample : public VulkanExampleBase
...
 				}
-
+				readSet = 1 - readSet;
 			}
 
-			// release the storage buffers back to the graphics queue
-			addComputeToGraphicsBarriers(compute.commandBuffers[i], VK_ACCESS_SHADER_WRITE_BIT, 0, VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT, VK_PIPELINE_STAGE_BOTTOM_OF_PIPE_BIT);
+			// The last iteration wrote to the "input" buffer, so we need copy the results to the output buffer
+			// First we need to put the buffers into transfer mode
+			std::array<VkBufferMemoryBarrier, 2> bufferBarriers;
```
jherico, line 325: Because the last dispatch is odd, it wrote to `input`, so we want to copy to `output`, so we need a barrier for the copy

examples/computecloth/computecloth.cpp
```diff
@@ -306,26 +300,66 @@ class VulkanExample : public VulkanExampleBase
...
+				0, nullptr);
+
+			// Then we copy the data
+			VkBufferCopy copyRegion{};
+			copyRegion.size = storageBuffers.output.size;
+			vkCmdCopyBuffer(compute.commandBuffers[i], storageBuffers.input.buffer, storageBuffers.output.buffer, 1, &copyRegion);
+
+			// Finally, move the input buffer back to it's original state as a read only buffer
+			// The output buffer has it's own barrier to move it to the graphics queue below
+			bufferBarriers[0].srcAccessMask = VK_ACCESS_TRANSFER_READ_BIT;
```
jherico, line 350: and finally a barrier for the input to move back to the expected starting state.

base/VulkanInitializers.hpp
jherico: I can revert this change and make sure all my new barriers explicitly set it then.

examples/computecloth/computecloth.cpp
```diff
@@ -365,13 +399,14 @@ class VulkanExample : public VulkanExampleBase
 			storageBufferSize,
 			particleBuffer.data());
 
-		// SSBOs will be used both as storage buffers (compute) and vertex buffers (graphics)
+		// storageBuffers.input is used as an SSBO and only on the compute queue as an intermediate location for iterative computation
 		vulkanDevice->createBuffer(
-			VK_BUFFER_USAGE_VERTEX_BUFFER_BIT | VK_BUFFER_USAGE_STORAGE_BUFFER_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT,
+			VK_BUFFER_USAGE_STORAGE_BUFFER_BIT | VK_BUFFER_USAGE_TRANSFER_DST_BIT | VK_BUFFER_USAGE_TRANSFER_SRC_BIT,
```
jherico, line 404: As noted, `storageBuffers.input` is never accessed on the graphics queue, so no need for the vertex buffer bit, but I do need to copy from it to the output buffer in the compute queue, so added the transfer src bit.

examples/computecloth/computecloth.cpp
```diff
@@ -382,11 +417,32 @@ class VulkanExample : public VulkanExampleBase
 		VkCommandBuffer copyCmd = vulkanDevice->createCommandBuffer(VK_COMMAND_BUFFER_LEVEL_PRIMARY, true);
 		VkBufferCopy copyRegion = {};
 		copyRegion.size = storageBufferSize;
+		vkCmdCopyBuffer(copyCmd, stagingBuffer.buffer, storageBuffers.input.buffer, 1, &copyRegion);
```
jherico, line 420: So this dual copy is a little weird.  

Basically we want the input buffer to contain the data at the beginning of the compute stage, so we populate it here, but we also want the output buffer to be owned by the graphics queue initially, so we also populate it here. 
Technically we don't really need to populate it, we just need to have it accessed by the graphics queue in some way to take initial ownership, so the initial release barrier pairs up.

examples/computecloth/computecloth.cpp
```diff
@@ -595,14 +651,35 @@ class VulkanExample : public VulkanExampleBase
...
+
+		VkBufferMemoryBarrier bufferBarrier = vks::initializers::bufferMemoryBarrier();
+		bufferBarrier.srcAccessMask = 0;
+		bufferBarrier.dstAccessMask = VK_ACCESS_SHADER_READ_BIT;
+		bufferBarrier.srcQueueFamilyIndex = vulkanDevice->queueFamilyIndices.graphics;
+		bufferBarrier.dstQueueFamilyIndex = vulkanDevice->queueFamilyIndices.compute;
+		bufferBarrier.size = VK_WHOLE_SIZE;
+		bufferBarrier.buffer = storageBuffers.input.buffer;
+		// Explicitly acquire the storageBuffer.input buffer (the output buffer will ping-pong back and forth so it's acquired as part of the main compute command buffer)
+		vkCmdPipelineBarrier(acquireCmd, VK_PIPELINE_STAGE_TOP_OF_PIPE_BIT, VK_PIPELINE_STAGE_COMPUTE_SHADER_BIT, VK_FLAGS_NONE, 0, nullptr, 1, &bufferBarrier, 0, nullptr);
```
jherico, line 679: this queue transition could be eliminated if we deferred populating the input buffer with the initial contents until here and did it explicitly on the compute queue.  That would probably be cleaner since we'd no longer ever need to move the input buffer across queues, but we'd still need to explicitly have the graphics queue take ownership of the output buffer and release it to the compute queue.

Can you have a queue take ownership by creating a command buffer that does _nothing_ but execute a release pipeline barrier?

base/VulkanDevice.cpp
jherico: I'd like to contribute more to the official repo, but I find working with the C API excruciating after working with the C++ API for so long.

examples/computecloth/computecloth.cpp
```diff
@@ -306,26 +303,68 @@ class VulkanExample : public VulkanExampleBase
 
 			// Dispatch the compute job
 			const uint32_t iterations = 64;
+			static_assert(iterations % 2 == 0, "The below code assumes an even number of iterations.");
+			uint32_t readSet{ 0 };
 			for (uint32_t j = 0; j < iterations; j++) {
-				readSet = 1 - readSet;
```
jherico, line 310: OK, I get the two command buffers now.  This was previously a global variable and because this is at the start of the loop, on the first iteration we read from the output buffer, which is what's populated by the original `prepareStorageBuffers`.  The second command buffer however would invert that.

However, since `readSet` is never modified outside of this function, during the actual render loop we only ever use one of the two command buffers.

I think having one compute command buffer and having it always treat the input and output buffers in the same way each frame is less error prone.

examples/computecloth/computecloth.cpp
jherico: I was initially avoiding this because I thought the resulting diff would look much larger because of moved code, but realized I can avoid moves by factoring out the particle vector computation to its own function.

jherico: OK, the diff is longer than I'd like, but the input buffer only ever gets used on the compute queue now.

PR #1114, opened 2024-03-23
Remove redefined variable in same scope (cascaded shadow mapping sample)

ntsh-oni:
The variable `i` is re-defined. While this compiles fine, it reduces code clarity.


SaschaWillems:

Issue #1004, opened 2022-12-19
rayquery example requires VK_KHR_ray_pipeline


bylaws:
Some HW, notably Snapdragon 8 Gen 2 devices support only VK_KHR_ray_query. As it is the ray query example ends up requiring support for ray pipeline at init time even if it's not used at runtime. Ideally this requirement could be relaxed to allow for using the example on such devices. 

(Would also be nice to have android support for rayquery since it's currently missing from the android/examples folder)

SaschaWillems:
I just checked, and the the ray tracing pipeline isn't enabled for the ray query sample:

```	if (!rayQueryOnly) {
		enabledDeviceExtensions.push_back(VK_KHR_RAY_TRACING_PIPELINE_EXTENSION_NAME);
	}
```

I'll add the missing android build files. Somehow missed this when the first devices came out that supported ray queries.

PR #1115, opened 2024-04-02
Remove rayTracingPipeline requirement from rayquery

werman:
rayquery test doesn't use ray tracing pipelines so it shouldn't enable this feature.

See: https://github.com/SaschaWillems/Vulkan/issues/1004


SaschaWillems:

PR #1117, opened 2024-04-23
macOS/iOS fixes plus other generic fixes for clang and validation warnings

SRSaunders:
This fixes a recently reported macOS command line build issue against merged PR #965.  In addition this fixes some other macOS issues and adds generic fixes for clang warnings and some validation layer warnings in a few examples:

1. Fixes macOS command line builds when using cmake's make and ninja generators (restore -ObjC++ compiler flag).
2. Updates macOS versions of `getAssetPath()` and `getShaderBathPath()` to be tolerant of command line and Xcode builds with associated binary locations (e.g. in **/bin** or **/bin/Debug** or **/bin/Release**).
3. Fixes `debugUtilsMessageCallback()` to be tolerant of cases when `pMessageIdName` is null, which can occur when validation is on and MoltenVK issues errors `[mvk-error]` or warnings `[mvk-warn]` such as using deprecated `VK_USE_PLATFORM_IOS_MVK` or `VK_USE_PLATFORM_MACOS_MVK` defines.
4. Fixes clang type warning in `getValueAsInt()` and missing overrides in the _variablerateshading_ example.
5. Fixes validation layer warning on exit in the _computeraytracing_ example - was missing the staging buffer destroy.
6. Restores the spacebar text visibility toggle in the _textOverlay_ example.  This must have been lost during refactoring.
7. Adds support for `VK_USE_PLATFORM_METAL_EXT` which supercedes the deprecated `VK_USE_PLATFORM_IOS_MVK` and `VK_USE_PLATFORM_MACOS_MVK` platform defines.  This touches a lot of files but the changes are straight forward.  The only issue is that `VK_USE_PLATFORM_METAL_EXT` requires initialization using a Metal layer vs. the view, so I had to make a few Apple-only changes to support this.  I retained the deprecated platform options for backwards compatibility, but the default cmake setting is now `VK_USE_PLATFORM_METAL_EXT`.
8. Check for supported dynamic state features in the _dynamicstate_ example vs. just the presence of extensions since these are not necessarily the same thing.  This is especially true for `VK_EXT_extended_dynamic_state3`.
9. Fixed the `vkCmdDraw()` `vertexCount` argument in the _particlesystem_ example since the previous code used the particle buffer size and not the particle count.  This lead to multiple drawing artifacts as would be expected.
10. Updated the examples list and restored the benchmarking script to the top level.  This must have been lost when removing the bin directory.  I still find the script useful when regression testing changes - it's a quick way to run through all the examples (with validation on) without manually running each one.
11. Fixed validation warning in _descriptorindexing_ example.  Was missing the `VK_KHR_maintenance1` extension.
12. Added check for device max pipeline recursion depth to avoid validation warnings for limited capability devices in the _raytracingcallable_, _raytracingintersection_, _raytracingreflections_, and _raytracingshadows_ examples.
13. Fixed build settings for OpenMP in the _texture3d_ example for all platforms.  This used to work, but somehow things must have changed and the compiler flag settings and library location were not being set for the target.
14. Updated and simplified build instructions for macOS.
15. Updated CI script to account for updated build instructions re libomp on macOS.
16. Added fix for combined image descriptor offset calculation in _descriptorbuffer_ example (#1118).
17. Add support for non-system level Vulkan SDK installs on macOS, with fallback to direct linkage to the MoltenVK library if the Vulkan loader is not found.  This is helpful when testing against different official SDK versions, or locally built versions of MoltenVK.  You can point `VULKAN_SDK` at a local install and cmake will use it if the system global installation is not available.

Tested on Windows 10, Manjaro Arch Linux, and macOS Ventura 13.6.6


SRSaunders:
The CI script is failing for macOS since the **libomp** path should now be `/usr/local/opt/libomp/lib/libomp.dylib` for homebrew on **x86_64**.   See the updated (simpler) build instructions for macOS in commit https://github.com/SaschaWillems/Vulkan/pull/1117/commits/20570645b180d1c07106b2218713702c60d152c7.

Even though it was building before, OpenMP was not actually working for any platform: Win, Linux, or macOS.

SaschaWillems:
Thank you very much for your PR. This is very much appreciated.

I will take a look at this, but it may take a few weeks.

SRSaunders:
Thanks for agreeing to look at this.  No rush so feel free to take your time.

There are a lot of small things collected in this PR.  Each commit is a separate item (except `VK_USE_PLATFORM_METAL_EXT`, which is a bit larger), so hopefully this will make it easier to review.

SRSaunders:
Thanks for merging.

PR #1121, opened 2024-05-11
Fix SSAO_NOISE_DIM for android

Vovan675:
Comment states that it should be "smaller", but its the opposite I think


SaschaWillems:

PR #1122, opened 2024-05-14
Fix SSAO blur range

Vovan675:
I think blur range should include last iteration. Otherwise it will be weight shifted by one texel.


SaschaWillems:
Yeah, that's correct indeed. But updating the text shaders only doesn't suffice. The samples consume the combined .spv files, so they need to be updated too.

PR #1125, opened 2024-05-15
Fix computeshader sample first frame being rendered with bogus UBOs

takayhan-AMD:
After this change, the sample's render() method would first update its UBO and then start drawing stuff so as to have valid frame output from frame 0, similarly to other samples.


SaschaWillems:
Thank you very much for this PR. Very much appreciated!

PR #1126, opened 2024-05-15
Compile all shaders and count errors

Vovan675:
Try to compile as many shaders as possible. If any of them are not compiled, just output error and skip it. At the end, print out how many errors occurred.
This makes life easier if you have some testing in one shader and broke it, but want to compile another one. Or, if you don't have glslang compiler that supports ray tracing extensions, it will just skip it and compile all other shaders.
Without this fix, I need to remove ray tracing shaders because my glslang compiler doesn't support it, otherwise I won't be able to compile SSAO shaders.

### DIFF
--- a/examples/ssao/ssao.cpp
+++ b/examples/ssao/ssao.cpp
@@ -14,9 +14,9 @@
 
 // We use a smaller noise kernel size on Android due to lower computational power
 #if defined(__ANDROID__)
-#define SSAO_NOISE_DIM 8
-#else
 #define SSAO_NOISE_DIM 4
+#else
+#define SSAO_NOISE_DIM 8
 #endif
 
 class VulkanExample : public VulkanExampleBase
--- a/shaders/glsl/compileshaders.py
+++ b/shaders/glsl/compileshaders.py
@@ -30,6 +30,7 @@ def findGlslang():
 glslang_path = findGlslang()
 dir_path = os.path.dirname(os.path.realpath(__file__))
 dir_path = dir_path.replace('\\', '/')
+errors_count = 0
 for root, dirs, files in os.walk(dir_path):
     for file in files:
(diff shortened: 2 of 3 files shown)


joshohagan:
You can clone my forked repository which compiles shaders using CMake on the master branch (much easier than using a python script): https://github.com/joshohagan/VulkanSamples/tree/master

SaschaWillems:
>  if you don't have glslang compiler that supports ray tracing extensions

What compiler is that? glslangvalidator has support for ray tracing and I'm not aware of any recent version that doesn't support the RT shader types.

SaschaWillems:
The one from the SDK works fine with ray tracing. If a shader files running this script, it's mostly a problem with the script. If you post the error you get, I'll take a look.

Vovan675:
When I am running glsl's compileshares.py I get this. Maybe we need to up target spirv version or something. But handling these errors and don't stop on them I think more desirable than just stopping compiling other shaders)
```
F:/Repositories/Vulkan/shaders/glsl\rayquery\scene.frag
ERROR: F:/Repositories/Vulkan/shaders/glsl\rayquery\scene.frag:2: 'GL_EXT_ray_tracing' : not supported for current targeted SPIR-V version
ERROR: F:/Repositories/Vulkan/shaders/glsl\rayquery\scene.frag:2: '#extension' : extra tokens -- expected newline
ERROR: F:/Repositories/Vulkan/shaders/glsl\rayquery\scene.frag:2: '' : compilation terminated
ERROR: 3 compilation errors.  No code generated.
```

SaschaWillems:
That's the script's fault. This is for inline ray tracing, which is using the fragment shader stage, so the rt stage handling doesn't work there. I'll take a look at that.

SaschaWillems:
I have pushed a fix for both the GLSL and HLSL compilation scripts. Both should now be able to compile all shaders using the compilers from the SDK.

PR #1127, opened 2024-05-16
Apply a random seed when NOT in benchmark mode to have 100% deterministic runs in all samples

takayhan-AMD:
These samples lack the check for benchmark.active when applying a random seed, which is done for other samples.

Tested locally, all samples continue to work fine as expected, in either mode.


SaschaWillems:

PR #1132, opened 2024-05-24
Add CMake shader compilation

joshohagan:
The problems with the current shader compilation are:
- Some ray tracing shaders cannot compile by default, and their compilation failures cause no more shaders afterwards to be compiled.
- The current process has to be run by a python script and all shaders are recompiled, which is slow.

This MR addresses these issues by cherry-picking the cmake shader compilation commit from the "cmake_shaders" branch.

P.S. Please accept this ASAP Sascha, it is a really annoying issue for people like me who want to compile shaders in your repository for my job.

Also don't forget to delete the "cmake_shaders" branch once this MR is accepted.


SaschaWillems:
> P.S. Please accept this ASAP Sascha, it is a really annoying issue for people like me who want to compile shaders in your repository for my job.

Please remember that I'm doing all this in my spare time, and I don't like to be pressured. And what job does you require to compile my samples? Are you working for an IHV?

joshohagan:
Hi Sascha,

Sorry, I didn't mean to pressure you. I didn't realize that you are doing this in your spare time, so that's fair enough that you don't want to have to work on the repository very often. On second thoughts, this isn't very urgent anyway as if people need to compile shaders they can download my forked repo.

For needing the shaders for my job, I was referring to my game dev company wanting to compile the shaders to play around with the code, but on second thoughts that information isn't very relevant.

shaders/hlsl/README.md
```diff
@@ -1,6 +1,6 @@
 ## HLSL Shaders
 
-This directory contains shaders using the [HLSL shading language](https://docs.vulkan.org/guide/latest/hlsl.html). These can be compiled with [DXC](https://github.com/microsoft/DirectXShaderCompiler) using e.g. the included `compile.py` script.
+This directory contains shaders using the [HLSL shading language](https://docs.vulkan.org/guide/latest/hlsl.html). To use the HLSL shaders instead of GLSL, change the line `std::string shaderDir = "glsl";` in `vulkanexamplebase.h`. These shaders are compiled with [DXC](https://github.com/microsoft/DirectXShaderCompiler) using CMake when you make an edit to the shader file.
```
SaschaWillems, line 3: That's not correct. There is no need to change code to switch between shader types. That's done via a command line argument, see the main readme.

So this needs to be reverted.

shaders/hlsl/compile.py
SaschaWillems, line 1: Please undo this.

shaders/glsl/compileshaders.py
SaschaWillems, line 1: Please undo this.

examples/CMakeLists.txt
```diff
@@ -83,6 +124,10 @@ function(buildExamples)
 	endforeach(EXAMPLE)
 endfunction(buildExamples)
 
+# Get shader compilers from SDK
+set(GLSL_VALIDATOR "$ENV{VULKAN_SDK}/Bin/glslangValidator.exe")
```
SaschaWillems, line 128: Don't explicitly set these, there are people use those compiler from different paths.

examples/CMakeLists.txt
```diff
@@ -41,6 +41,47 @@ function(buildExample EXAMPLE_NAME)
...
+			set(PROFILE "vs_6_1")
+		elseif(${FILE_EXT} STREQUAL ".frag")
+			set(PROFILE "ps_6_1")
+		elseif(${FILE_EXT} STREQUAL ".comp")
+			set(PROFILE "cs_6_1")
+		elseif(${FILE_EXT} STREQUAL ".geom")
+			set(PROFILE "gs_6_1")
+		elseif(${FILE_EXT} STREQUAL ".tesc")
+			set(PROFILE "hs_6_1")
+		elseif(${FILE_EXT} STREQUAL ".tese")
```
SaschaWillems, line 69: This is missing several stages used in this repo.

SaschaWillems (changes requested):
I did a first quick review and this needs some changes. 

I'm also not sure if I really want shader compilation added via cmake. There is a reason I added the compiled SPIR-V files, so only maintainers need to worry about compiliation.

And what will happen if someone tries to run CMake (or a build) that doesn't have dxc/glslang? Will that fail? Will that work? Will this always overwrite all SPIR-V files at first compilation?

joshohagan:
Hi Sascha,

Sorry for the VERY late reply, I missed the notification for this and will check my notifications from now on.

I read your comment about not wanting CMake shader compilation in case people do not have dxc/glslang, and I agree. I think the best solution is for people to use the compile shaders script, which you fixed in a recent issue. If programmers want the compilation to be faster, they can always modify the script to only compile shaders when the file has been changed.

PR #1136, opened 2024-05-27
Gears, Switched Vertex Indexes From 32bit to 16bit

ahcox:
ARM FrameAdvisor showed 32bit indexes in use. Since there are only 880 triangles at present, 65536 vertices leaves room for growth. Keeping data types as narrow as possible is good practice for mobile where bandwith use => power => heat => thermal throttling and shortened runtimes. I think this change improves the example for a project used by people learning graphics for Android.

The mesh complexity metrics in frame advisor show the improvement:
gears_32_bit_indexes
gears_16_bit_indexes

### DIFF
--- a/examples/gears/gears.cpp
+++ b/examples/gears/gears.cpp
@@ -1,9 +1,9 @@
 /*
 * Vulkan Example - Drawing multiple animated gears (emulating the look of glxgears)
-* 
+*
 * All gears are using single index, vertex and uniform buffers to show the Vulkan best practices of keeping the no. of buffer/memory allocations to a mimimum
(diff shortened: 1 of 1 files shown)


SaschaWillems:
I'm kinda hesitant to merge (mostly) vendor specific changes, but in general I'm not against it. 

The repo already has a sample specific to vertex attributes at https://github.com/SaschaWillems/Vulkan/tree/master/examples/vertexattributes

That would probably the better sample to adjust, and add a vendor specific path incl. documentation.

ahcox:
Your advice on a better spot for it is taken.

PR #1137, opened 2024-05-27
Gears - 8bit Per-vertex Color

ahcox:
Reduced Vertex size from 36 to 28 bytes, a bandwidth, power, heat, and thus battery life and performance gain for mobile devices like Android phones. Here's the before and after in ARM FrameAdvisor:

Screenshot from 2024-05-27 16-04-17

gears_8bit_color annotated

Note the vertex color could be eliminated and replaced with a push_constants of the per-object color ahead of each draw command + modifications to the shaders, but this is less intrusive and a good step along the way.

### DIFF
--- a/examples/gears/gears.cpp
+++ b/examples/gears/gears.cpp
@@ -1,9 +1,9 @@
 /*
 * Vulkan Example - Drawing multiple animated gears (emulating the look of glxgears)
-* 
+*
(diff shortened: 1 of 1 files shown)


ahcox:
I feel this is similar to my other PRs. Perhaps not aligned with the repo goals. Not interested in wasting your time.

PR #1166, opened 2024-11-18
Adding-Command-Line-Argument-rp-for-Non-Mac-iOS-platforms

quic-sagrahar:
It helps in dynamically locating directory where shaders and assets folders present. Useful for cross compiled targets , where hard coded paths are based on source platforms which may non accessible on target platforms.

### DIFF
--- a/base/VulkanTools.cpp
+++ b/base/VulkanTools.cpp
@@ -12,6 +12,9 @@
 // iOS & macOS: getAssetPath() and getShaderBasePath() implemented externally for access to Obj-C++ path utilities
 const std::string getAssetPath()
 {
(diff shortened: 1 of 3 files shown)


base/VulkanTools.cpp
```diff
@@ -12,6 +12,9 @@
 // iOS & macOS: getAssetPath() and getShaderBasePath() implemented externally for access to Obj-C++ path utilities
 const std::string getAssetPath()
 {
+if(vks::tools::resourcePathSet){
```
SaschaWillems, line 15: You can just check if `resourcePath`

SaschaWillems:

Will do some testing first though.

PR #1172, opened 2024-12-16
Swappy and GameActivity

HakimHauston:
Integration to 2 of the projects:
-. Deferred Multi Sampling (base for the initial work)
-. GLTFSceneRendering (patched the work with diff and some tuning)

### DIFF
--- a/android/build.gradle
+++ b/android/build.gradle
@@ -40,8 +40,8 @@ allprojects {
 ext {
     abiFilters = "arm64-v8a"
     minSdkVersion = 21
-    targetSdkVersion = 26
-    compileSdkVersion = 26
+    targetSdkVersion = 35
+    compileSdkVersion = 35
     shaderPath = '../../../shaders/'
     assetPath = '../../../assets/'
 }
--- a/android/examples/base/CMakeLists.txt
+++ b/android/examples/base/CMakeLists.txt
@@ -9,6 +9,8 @@ include_directories(../../../external/gli)
 include_directories(../../../external/imgui)
 include_directories(${EXTERNAL_DIR}/tinygltf)
 include_directories(${ANDROID_NDK}/sources/android/native_app_glue)
+include_directories(${game-activity-include})
+include_directories(${games-frame-pacing-include})
 
 set(KTX_DIR ../../../external/ktx)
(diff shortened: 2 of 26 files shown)


HakimHauston:
Hi @SaschaWillems ,

Thanks for your amazing work on this Vulkan Sample repository!

I noticed that there are some possible performance gain for the Android platform, the largest of which can be gain through integrating [Swappy (Android's Frame Pacing API)](https://developer.android.com/games/sdk/frame-pacing#advantages) and [GameActivity](https://developer.android.com/stories/games/cat-daddy-agdk).

I've based my work on your [deferredmultisampling](https://github.com/SaschaWillems/Vulkan/tree/master/android/examples/deferredmultisampling) sample and plan to patch the rest of the projects. I've patched [gltfscenerendering](https://github.com/SaschaWillems/Vulkan/tree/master/android/examples/gltfscenerendering) as a separate commit cc275174 but figured some minor manual modification may be required for each of the projects.

Before I do the manual patch for each of the projects, would you please:
1. let me know if there is an easier way to patch the same changes to all `examples` ? I saw your [_template](https://github.com/SaschaWillems/Vulkan/tree/master/android/examples/_template) directory but looks like it is for creating a new sample rather than patching existing ones
2. review my modification for `deferredmultisampling` at commit 6e58227e to ensure that it is acceptable? I'll base the patch for the rest of the examples on this commit.

SaschaWillems:
I'll probably need some more time for a proper review, but in general changes like this need to work an all platforms, so changes like in https://github.com/SaschaWillems/Vulkan/pull/1172/files#diff-5b12913c531aefa5a62f2e7cf59ce9479ee5491c67cca7c898baa8d45f38b960L45 need to be properly guarded so they only apply to Adnroid without touch other platforms.

And if you need to have changes applied to all or many samples, such changes need to be implemented in the example base class..

I also need to read up on swappy first. It looks like integrating that would require a lot of changes and I'm not sure about that. Who would maintain that part of the code in the futurte?

SaschaWillems:
Any update on this?

SaschaWillems:
Closing this due to inactivity. If you want to pursue this further, feel free to open an issue for further disucssions.

PR #1174, opened 2025-01-01
Update LICENSE.md, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛


SaschaWillems:
There is no need to update the copyright year in the license. Also simply changing the year, instead of making it a range is wrong.

PR #1195, opened 2025-03-04
Mediacodec works in portrait

zhuker:


zhuker:
wrong origin again :)

PR #1202, opened 2025-04-28
Remove redundant fields assignments in VkSamplerCreateInfo

Nomyo:
Simple clean-up in sampler creation


SaschaWillems:
No longer applies.

PR #1209, opened 2025-06-05
Fix Validation Log to display sample title (name) vs. engine name

### LINKED ISSUE #1208: Validation Log displays generic engine name vs. sample name
The validation log (-vl) shows "Sample: vulkanExample" for every sample that is run.  This makes it hard to figure out which validation warnings/errors correspond to which sample, especially when running a batched sequence.

Instead, the log should record the sample title (vs. the engine name).

I will submit a PR to fix.

SRSaunders:
Fixes #1208.

In addition this PR makes the _*headless_ examples App and Engine names consistent with the project.

Also updates the title for _texturecubemaparray_ to make it unique in the log (vs. _texturecubemap_).


SaschaWillems:
Well, I just recently added that for a very specific reason. Still have uncommited changes on my disk.

SRSaunders:
No worries.  It looked like a useful capability, but if you have something else in mind then please just  close this PR and the original issue.

PR #1219, opened 2025-07-02
Make triangle example use base class logic for getting entrypoint name

LegNeato:
This is needed as in my rust fork we modify the logic in the base class, but there is nothing rust-specific about this so it should probably go upstream.


SaschaWillems:
Can you elaborate on the change. I'd like to keep my samples clean and easy to follow with as little indirection as required. So if this is a more fundamental change I might decide to not merge it.

PR #1220, opened 2025-07-02
Add screenshot capture to benchmark system for automated shader testing

LegNeato:
- Add `saveScreenshot()` method to capture swapchain images as PPM files
- Extend benchmark class to support screenshot capture at specified frame
- Add command-line flags `--screenshot` and `--screenshotframe`

This enables automated differential testing (not included in this PR, but trivial to add on top. I am doing so in our Rust fork)

### DIFF
--- a/base/benchmark.hpp
+++ b/base/benchmark.hpp
@@ -28,11 +28,13 @@ namespace vks
 		uint32_t duration = 10;
 		std::vector<double> frameTimes;
 		std::string filename = "";
+		std::string screenshotPath = "";
+		int32_t screenshotFrameIndex = -1;
 
 		double runtime = 0.0;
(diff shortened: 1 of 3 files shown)


LegNeato:
(note I am not a C++ programmer so feel free to give pointers and corrections)

SaschaWillems:
Will first need to test if this actually works. I did something similar a few years ago, but ran into issues with sync on some platforms/devices. 

No need for the differential testing script from my side though.

LegNeato:
Sounds good! Looks the differential testing isn't that useful on a lot of the output anyway due to float differences. I only tested this on mac with MoltenVK FWIW (which means some shaders were not tested of course).

PR #1231, opened 2025-08-15
Reducing draw calls in textoverlay example

GDBobby:
if this pull request is denied, I'd still recommend changing line 107 to

VkDeviceSize bufferSize = TEXTOVERLAY_MAX_CHAR_COUNT * sizeof(glm::vec4) * 4;

or renaming TEXTOVERLAY_MAX_CHAR_COUNT to TEXTOVERLAY_MAX_VERTEX_COUNT

addText should have protections for vertexBuffer overflow (situations in which numLetters >= TEXTOVERLAY_MAX_CHAR_COUNT), but I'm assuming that's out of scope for this example.

there's still room for other easy optimizations. 
like an index buffer, 
or a separate buffer containing font data as a whole, reducing uploaded data and cpu calculations.

### DIFF
--- a/examples/textoverlay/textoverlay.cpp
+++ b/examples/textoverlay/textoverlay.cpp
@@ -104,7 +104,7 @@ struct TextOverlay
 		stb_font_consolas_24_latin1(stbFontData, font24pixels, fontHeight);
 
 		// Vertex buffers containing the text information per max. concurrent frames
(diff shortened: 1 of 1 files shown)


SaschaWillems:
Thanks for your PR. Your contribution is very much appreciated.

As for further improvements: I want to keep this sample as simple as possible, it's kinda like a precursor to the ImGui sample. So I'm fine with the way it is (after this PR).

PR #1238, opened 2025-09-30
Fix incorrect uniform buffer index used to bind the descriptor buffer at cmd buffer building time

takayhan-AMD:
Trivial fix to render using correct UBO index.


SaschaWillems:
Thx for this PR. Very much appreciated!

PR #1239, opened 2025-10-03
Fix textoverlay Android build - include torusknot.gltf asset

lordalcol:
Fixes a problem where it wouldn't find the torusknot.gltf model on android.


SaschaWillems:

PR #1241, opened 2025-10-22
Fix instance creation on apple platforms

aitor-lunarg:
Only MoltenVK requires configuration at startup

### DIFF
--- a/base/vulkanexamplebase.cpp
+++ b/base/vulkanexamplebase.cpp
@@ -837,20 +837,36 @@ VulkanExampleBase::VulkanExampleBase()
 		vks::tools::resourcePath = commandLineParser.getValueAsString("resourcepath", "");
 	}
 #else
-	// On Apple platforms, use layer settings extension to configure MoltenVK with common project config settings
-	enabledInstanceExtensions.push_back(VK_EXT_LAYER_SETTINGS_EXTENSION_NAME);
+	// With MoltenVK, use layer settings extension to configure it with common project config settings
+	// Other implementations like lavapipe and KosmicKrisp do not need this
+	uint32_t extCount = 0;
+	vkEnumerateInstanceExtensionProperties(nullptr, &extCount, nullptr);
+	if (extCount > 0)
+	{
+		std::vector<VkExtensionProperties> extensions(extCount);
+		if (vkEnumerateInstanceExtensionProperties(nullptr, &extCount, &extensions.front()) == VK_SUCCESS)
+		{
+			for (VkExtensionProperties& extension : extensions)
+			{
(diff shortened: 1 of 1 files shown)


SaschaWillems:

SRSaunders:
Thanks @spencer-lunarg and @aitor-lunarg for your answers above.  In conclusion I recommend:

1. For _debugprintf_: Remove the `VK_EXT_layer_settings` extension since it not checked. The remainder of the layer settings code should remain in place so the VVL can enable shader `debugPrintfEXT` and `printf_to_stdout` on macOS.  On other platforms shader `debugPrintfEXT` should work natively in conjunction with RenderDoc.  As far as I know, RenderDoc is not available on macOS and therefore `printf_to_stdout` is required there as an alternative, no matter what driver is used (MoltenVK or KosmicKrisp).
2. For _descriptorindexing_: Check if `VK_EXT_layer_settings` is available in the driver (i.e. MoltenVK) and only enable MoltenVK's `MVK_CONFIG_USE_METAL_ARGUMENT_BUFFERS` setting if so.  However, this whole thing is only for older versions of the Apple SDK (and MoltenVK for macOS and iOS) where argument buffers were not turned on by default.  With the newest SDKs, MoltenVK argument buffers are enabled by default in macOS and iOS and the layer setting only serves to assert this - a noop really.

Both of the above changes should get out of the way of KosmicKrisp.  I would be happy to submit a PR for this.

> It's up to the user to set the environment variable for MoltenVK to expose the required utility for the sample to be supported (I recall MoltenVK has such a thing). So, trying to configure MoltenVK through VK_EXT_layer_settings seems like a bad practice to me. This should be done through environment variables before running the sample.

Both the VVL and MoltenVK offer configuration options by environment variable and via program control.  In addition the VVL has the configurator, which is quite powerful and automates various use cases via a GUI.  I believe both env variables and program control should be offered for end user flexibility.  Regarding the two samples in question here, when first making them work on macOS and iOS I chose program control since they would then work out of the box.  Depending on an end-user setting an environment variable (VVL or MoltenVK) is quite brittle - they may or may not read the docs.  It also embeds instructional information in the code which alerts the programmer regarding any dependencies on Apple platforms.  I recognize that KosmicKrisp seeks to eliminate those dependencies going forward, but that has been the reality with the Apple platform to date.

spencer-lunarg:
> MoltenVK will ignore its own layer settings in VkLayerSettingsCreateInfoEXT unless VK_EXT_layer_settings is enabled for the driver.

wow... well it a fair choice for them to do that according to the spec... I have zero idea **why** they would want to, seems like a way to confuse people as there is no way to validate this currently

PR #1243, opened 2025-10-29
Layer settings compatibilty fixes for debugprintf and descriptorindexing samples

SRSaunders:
See discussion in #1241.

This PR should improve compatibility with KosmicKrisp for samples _debugprintf_ and _descriptorindexing_.  The **VK_EXT_layer_settings** extension should now be enabled and used only for MoltenVK-specific cases (i.e. _descriptorindexing_).  Layer settings for the VVL do not require the extension (i.e. _debugprintf_).

@aitor-lunarg please review this PR and test to make sure these changes work with KK.  Thanks.

One more thing:  I may be able to remove all specializations for MoltenVK in the _descriptorindexing_ sample if we can require a **minimum SDK version** on Apple.  It seems the variable descriptor count issue with MoltenVK may have been solved starting in Vulkan SDK 1.4.321.0.  And MoltenVK argument buffers have been enabled by default prior to this SDK version as well.  I would like to test this a bit more before drawing final conclusions, but I may add an additional commit here if it proves out.  This would remove Apple-specific code paths within the _descriptorindexing_ sample and clean things up.

### DIFF
--- a/examples/descriptorindexing/descriptorindexing.cpp
+++ b/examples/descriptorindexing/descriptorindexing.cpp
@@ -234,28 +234,6 @@ public:
 		// Binding 1 are the fragment shader images, which use indexing
 		// In the fragment shader:
 		//	layout (set = 0, binding = 1) uniform sampler2D textures[];
-
-#if (defined(VK_USE_PLATFORM_IOS_MVK) || defined(VK_USE_PLATFORM_MACOS_MVK) || defined(VK_USE_PLATFORM_METAL_EXT))
-		// Disable variable descriptor count feature on macOS/iOS until MoltenVK supports this feature when using combined image sampler textures
-		// Other implementations like lavapipe and KosmicKrisp do not need this, but retain for compatibility (for now) across drivers on macOS & iOS
-		// Note we are using only 1 descriptor set with a fixed descriptor count/pool size, so we can simply turn off the capability for now
-		std::vector<VkDescriptorBindingFlagsEXT> descriptorBindingFlags = {
-			0,
-			0
-		};
-		setLayoutBindingFlags.pBindingFlags = descriptorBindingFlags.data();
-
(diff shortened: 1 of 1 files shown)


SRSaunders:
Thanks @aitor-lunarg for your review.  In the end I have removed the MoltenVK layer setting for the _descriptorindexing_ sample.  This setting has not been required since SDK 1.3.296.0 (released Oct 8, 2024) which is over 1 year now.  Starting with this version MoltenVK has been configured to use Metal argument buffers by default and the layer setting was redundant.  I believe this is a long enough period to proceed with this change, removing older SDK version legacy support.

@SaschaWillems I would like to ask a question regarding your project's policy regarding SDK version compatibility.  Do you recommend only the latest SDKs, or do you have a back version compatibility policy?  The reason I ask is that remaining code differences on Apple platforms (using MoltenVK up to now) can be almost fully eliminated for the _descriptorindexing_ sample starting with SDK 1.4.321.0 (released July 15, 2025).  A defect in MoltenVK that was preventing use of variable descriptor counts was corrected in that SDK.  However, this may be a bit too recent and I am hesitant to proceed with this change until I understand your project's SDK back version policy.  Would you please provide some guidance here.  Thanks.

SaschaWillems:
I do not have an SDK policy except for "you don't need the SDK to use my samples". I'm fine with things only working with the most recent SDK. So feel free to adjust accordingly ;)

SRSaunders:
Thanks @SaschaWillems for your guidance here.  I have restored the _descriptorindexing_ sample to use variable descriptor counts on macOS & iOS as per the original code.  The only Apple-specific thing remaining in the sample is usage of `UPDATE_AFTER_BIND` flags to increase the maximum number of descriptors per-stage.  These are needed to silence VVL warnings about limits being exceeded, at least on MoltenVK.  I am not sure what limits KosmicKrisp will have, but usage of these flags is pretty benign and should work regardless of the driver (even on Windows or Linux for that matter).

The PR is now ready for your review.

PR #1246, opened 2025-11-28
Add single-sample shader compilation to GLSL/HLSL script

takayhan-AMD:
It's time-consuming to have to wait for all shaders of all samples to get compiled, so I replicated the behavior already in the slang script.

Tested it locally and it seems to work as expected.


SaschaWillems:
Thanks for your contribution. Very much appreciated.

PR #1249, opened 2025-12-26
shadowmapping: Use current frame's shadow matrix in scene pass

aitor-lunarg:
Using previous frame's shadow matrix leads to visual artifacts since the transformation will not be the same as the one used to generate the shadow map.


SaschaWillems:
Thanks. Very much appreciated 👍🏻

PR #1259, opened 2026-03-05
Compile and release Windows binaries

ThreeDeeJay:
Builds and uploads artifacts on every commit push to:
- GitHub Actions workflow 
  - https://github.com/ThreeDeeJay/Vulkan/actions/runs/22709555849
- Release - Latest
  - https://github.com/ThreeDeeJay/Vulkan/releases/tag/latest-master
  - https://github.com/ThreeDeeJay/Vulkan/releases/download/latest-master/Vulkan.Samples.zip
- Release - Archive
  - https://github.com/ThreeDeeJay/Vulkan/releases/tag/archive-master
  - https://github.com/ThreeDeeJay/Vulkan/releases/download/archive-master/Vulkan.Samples_r2906@d4bb3db.zip


ThreeDeeJay:
It's just a simple way to download from a static, permanent location without requiring and account (none of which is possible when just uploading binaries to the workflow), especially for those who want to quickly test specific Vulkan features but aren't familiar with compiling source code. Not to mention, it's way more practical from a developer's standpoint so they don't have to compile on their own machine, avoiding unneeded hassle, or even the privacy concerns of their username being leaked into the binaries that contain a path to `C:\Users\...`

As for the archive tag, it collects a copy of every single release which could help with tracking down regressions/debugging.
But what are the arguments against releasing Windows binaries, if I may ask? 🤔

Issue #1267, opened 2026-04-07
dynamicrenderinglocalread breaks on initial resize for device screens


SRSaunders:
The new _dynamicrenderinglocalread_ example breaks when starting on a non-default window size, for example when deployed to a physical device like an iPhone or the iOS Simulator.  It's likely an issue on Android too but I can't test that.

The failure is because the attachment resize logic within `setupFrameBuffer()` is called immediately by `VulkanExampleBase::prepare()` and thus `createAttachments()` and `vkUpdateDescriptorSets()` are called before the `vkCmdPipelineBarrier2KHR` function pointer is defined and before the descriptor sets are even allocated.

A small set of changes to the example's resize logic will make this work properly.  I will submit a PR to address it.

SaschaWillems:
I know. It's still work in progress.

SRSaunders:
I have a PR that fixes that part only.  Do you want me to post it?

SRSaunders:
I posted my proposed change as #1268.

@SaschaWillems please feel free to accept or disregard if you have your own solution.

SaschaWillems:
Should be fixed by https://github.com/SaschaWillems/Vulkan/commit/7bb35f007f29a25a90a93715abe6046d43473e9a, which works fine for me on all platforms I can test on

SRSaunders:
Thanks @SaschaWillems.  Your change works fine with iOS.

I have repurposed my PR to adding the example to the iOS subproject and making it work with the Simulator.

PR #1268, opened 2026-04-07
Fix colorAttachmentCount error in dynamicrenderinglocalread example

SRSaunders:
Fixes ~~#1267~~ #1271.

~~This change fixes startup resize for _dynamicrenderinglocalread_ to support non-default device screen sizes.~~

~~This also updates the iOS subproject to include this new sample and make it work on iOS devices and the Simulator.~~

**UPDATED**: Now fixes some validation layer errors in the example (all platforms):

1. ~~Wrong image layout validation error.~~ Original fix in this PR has now been reverted.  This remains an outstanding VVL bug, see https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/12039
2. Fixes the non-matching color attachment count when drawing the UI (1 actual vs. 4 expected)
3. ~~Failure to destroy buffer and memory on exit.~~ Already fixed in master.
4. Fixes some missing overrides (static analysis warnings in IDE)

### DIFF
--- a/examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
+++ b/examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
@@ -141,7 +141,7 @@ public:
 	{
 		// With VK_KHR_dynamic_rendering we no longer need a frame buffer, so skip the sample base framebuffer setup
 
-		if (attachments.width != width || attachments.height != height)
+		if (resized)
 		{
 			attachments.width = width;
 			attachments.height = height;
@@ -160,6 +160,7 @@ public:
 				writeDescriptorSets.push_back(vks::initializers::writeDescriptorSet(passes.composition.descriptorSet, VK_DESCRIPTOR_TYPE_INPUT_ATTACHMENT, static_cast<uint32_t>(i), &descriptorImageInfos[i]));
 			}
 			vkUpdateDescriptorSets(device, static_cast<uint32_t>(writeDescriptorSets.size()), writeDescriptorSets.data(), 0, nullptr);
+			resized = false;
 		}
 
 	}
@@ -458,6 +459,9 @@ public:
 		vkCmdEndRenderingKHR = reinterpret_cast<PFN_vkCmdEndRenderingKHR>(vkGetDeviceProcAddr(device, "vkCmdEndRenderingKHR"));
 		vkCmdPipelineBarrier2KHR = reinterpret_cast<PFN_vkCmdPipelineBarrier2KHR>(vkGetDeviceProcAddr(device, "vkCmdPipelineBarrier2KHR"));
 		vkCmdSetRenderingInputAttachmentIndicesKHR = reinterpret_cast<PFN_vkCmdSetRenderingInputAttachmentIndicesKHR>(vkGetDeviceProcAddr(device, "vkCmdSetRenderingInputAttachmentIndicesKHR"));
+
+		attachments.width = width;
+		attachments.height = height;
 
 		initLights();
 		createAttachments();
--- a/examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
+++ b/examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
@@ -103,8 +103,19 @@ public:
 		};
 		deviceCreatepNextChain = &enabledSynchronization2FeaturesKHR;
 
-		attachments.width = width;
-		attachments.height = height;
+#if (defined(VK_USE_PLATFORM_IOS_MVK) || defined(VK_USE_PLATFORM_METAL_EXT)) && TARGET_OS_SIMULATOR
+		static const VkBool32 valueFalse = VK_FALSE;
+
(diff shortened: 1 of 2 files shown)


SRSaunders:
FYI, another way to handle this could be to move the attachment resize logic out of `setupFrameBuffer()` and into a new `windowResized()` entry point override for the example.  Some examples do this, while others use the `resized` flag in either `setupFrameBuffer()` or `buildCommandBuffer()` (e.g. the ray tracing examples).

Given that it looks like a stylistic thing vs. functional, I decided to leave things alone and just use the `resized` flag.

SRSaunders:
Limited PR to adding _dynamicrenderinglocalread_ example to iOS subproject and iOS Simulator.

Resize logic now handled by changes within master branch.

SRSaunders:
Updated PR to include validation error fixes for _dynamicrenderinglocalread_.

examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
```diff
@@ -494,7 +510,7 @@ class VulkanExample : public VulkanExampleBase
 			colorAttachments[i] = {
 				.sType = VK_STRUCTURE_TYPE_RENDERING_ATTACHMENT_INFO,
 				.imageView = imageViewList[i],
-				.imageLayout = VK_IMAGE_LAYOUT_RENDERING_LOCAL_READ_KHR,
+				.imageLayout = (i == 0) ? VK_IMAGE_LAYOUT_COLOR_ATTACHMENT_OPTIMAL : VK_IMAGE_LAYOUT_RENDERING_LOCAL_READ_KHR,
```
SaschaWillems, line 498: I don't think this needs fixing. IMO this is a false positive and I've raised an issue in the VVL repo about this.

examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
```diff
@@ -101,6 +101,20 @@ class VulkanExample : public VulkanExampleBase
 			.synchronization2 = VK_TRUE,
 		};
 		deviceCreatepNextChain = &enabledSynchronization2FeaturesKHR;
+
+#if (defined(VK_USE_PLATFORM_IOS_MVK) || defined(VK_USE_PLATFORM_METAL_EXT)) && TARGET_OS_SIMULATOR
```
SaschaWillems: Isn't this something that should be fixed in MoltenVK. I'm a bit reluctant about including code that is so specific.

SRSaunders: I've tried to get this fixed for about 2 years.  See https://github.com/KhronosGroup/MoltenVK/issues/2285.  I gave up for the Vulkan-Samples project, and added this code for three of the samples there: _layout_transitions_, _pipeline_barriers_, and _subpasses_.  Vulkan-Samples will also require this setting for _dynamic_rendering_local_read_, but I have not submitted the change yet (but plan to soon since the iOS Simulator is documented as supported there).

However, I do understand your perspective on "keeping the samples simple and consistent" for your project in particular.  The issue is limited to the iOS Simulator, so running on macOS and iOS physical devices is fine without this change.  While there is Mac and iOS specific stuff in your base code, your samples by-and-large avoid that which is a good thing.  I can revoke this change if you like, especially since iOS (and Simulator) is more of a second class citizen for your project.  Please advise.

examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
SRSaunders: Interesting.  Referring to your issue https://github.com/KhronosGroup/Vulkan-ValidationLayers/issues/12039 where you discuss `VK_ATTACHMENT_UNUSED`, do you think this requires the `VK_EXT_dynamic_rendering_unused_attachments` extension to make this work for dynamic rendering?  

MoltenVK does not support this extension so I can't test there.  However, I will try on my Windows 11 and linux machines to see if my AMD card supports it, and report back later.

examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
SRSaunders: I deleted this section when merging in your lightsBuffer change.

examples/dynamicrenderinglocalread/dynamicrenderinglocalread.cpp
SRSaunders: Scratch that idea re the `VK_EXT_dynamic_rendering_unused_attachments ` extension.  Looks like that is for adapting to multiple pipelines where you can have an unused attachment in one, but not the other.

I will leave my original change in place for now and resolve once your VVL issue is looked at.

SRSaunders: @SaschaWillems I have given up waiting for the VVL team to fix this defect.  I have reverted my workaround in this PR and limited the scope to fixing the remaining `colorAttachmentCount` validation issue, as well as correcting a few IDE static analysis warnings.  See updated description at the top of this PR.

Should now be ready for your review and merge.

PR #1273, opened 2026-05-17
Add .rint compilation support to shaders/hlsl and shaders/glsl

nahjaeho:
Unlike shaders/slang/compileshaders.py, the scripts in shaders/hlsl and shaders/glsl did not support .rint compilation.

As a result, intersection.rint in the raytracingintersection sample was not compiled correctly by the GLSL shader compilation script.

This PR adds the missing .rint compilation support to the HLSL and GLSL shader scripts.


SaschaWillems:

PR #1274, opened 2026-05-25
Fix delayed UBO updates

takayhan-AMD:
This PR fixes two compute samples which invoke UBO update required for compute data too late before command buffer submission.

AFAICT the same pattern isn't found elsewhere now, but I have only taken a cursory look.


SaschaWillems:
Thanks. Very much appreciated!

PR #1280, opened 2026-07-13
Fix raytracinggltf GLSL shader code

tsymalla:
This app currently passes the payload from location 0 to TraceRayExt. However, payload slots 2 and 3 are never provided via TraceRayExt, so the AHS invoked during ray traversal will not be able to access the payload. In fact, this is correctly implemented in the Slang version of this example by using a wrapper struct.
The current GLSL version should show undefined behaviour, so it is a better and more safe approach to wrap all the payload types in a struct here as well.


SaschaWillems:
Thanks for fixing this. Very much appreciated!

PR #1286, opened 2026-08-17
Fix: shadowed fragments becoming transparent in rayquery example

deepaksinghdsk:
Bug in the rayquery example's shadow shader "shaders/glsl/rayquery/scene.frag":

if (shadowed) {
      outFragcolor *= 0.1; // scales alpha too, not just rgb
}

This scaled the whole vec4 including alpha, so shadowed pixels ended up with alpha = 0.1 instead of just a darker rgb. 
Went unnoticed on platforms where the swapchain picks **VK_COMPOSITE_ALPHA_OPAQUE_BIT_KHR** (alpha is ignored on present), but on WSI backends that don't support OPAQUE (e.g.Weston, which falls back to **PRE MULTIPLIED/POST_MULTIPLIED** "base/VulkanSwapChain.cpp:272"), the compositor blends using that alpha, making shadows show whatever is behind the window instead of just looking darker.

Fix: outFragcolor,rgb *= 0.1; Scale only rgb so alpha stays pinned at 1.0 regardless of which compositeAlpha mode gets selected.

Images of the final render in RenderDoc, with Alpha enabled and disabled for visualization of issue:
alpha_disabled alpha_enabled

### DIFF
--- a/shaders/glsl/rayquery/scene.frag
+++ b/shaders/glsl/rayquery/scene.frag
@@ -32,6 +32,6 @@ void main()
 
 	// If the intersection has hit a triangle, the fragment is shadowed
 	if (rayQueryGetIntersectionTypeEXT(rayQuery, true) == gl_RayQueryCommittedIntersectionTriangleEXT ) {
-		outFragColor *= 0.1;
+		outFragColor.rgb *= 0.1;
 	}
 }


deepaksinghdsk (commented):
Dev review done

